PR #49, opened 2019-01-10
Copy things instead of making other people own them

ColdenCullen:


source/client.c
```diff
@@ -491,12 +486,21 @@ int aws_mqtt_client_connection_connect(
...
     connection->state = AWS_MQTT_CLIENT_STATE_CONNECTING;
     connection->clean_session = clean_session;
     connection->keep_alive_time = keep_alive_time;
 
+    /* Cheat and set the tls_options host_name to our copy if they're the same */
+    if (tls_options) {
+        connection->tls_options = *tls_options;
+
+        const char *host_name = (const char *)connection->host_name->bytes;
+        if (0 == strcmp(host_name, connection->tls_options.server_name)) {
```
justinboswell: Isn't server_name the result of aws_tls_connection_options_set_server_name()? Is it set by this point? What will happen at clean up time since 2 things reference this stream now?

ColdenCullen: Yes, it is the result of that, and it should be set before this is called. The advantage of keeping tls options by copy now is that we know it can't outlife this connection, and as such, the host_name string can't outlife the tls_options.

justinboswell: Do we double free though?

ColdenCullen: WAIT NO HOLD ON, turns out it was being set unconditionally before. I'm going back to that model.

ColdenCullen: Also no, there is no double free, because tls options doesn't free anything.

include/aws/mqtt/client.h
```diff
@@ -239,8 +239,8 @@ int aws_mqtt_client_connection_connect(
     struct aws_mqtt_client_connection *connection,
     const struct aws_byte_cursor *host_name,
     uint16_t port,
-    struct aws_socket_options *socket_options,
-    struct aws_tls_connection_options *tls_options,
+    const struct aws_socket_options *socket_options,
+    const struct aws_tls_connection_options *tls_options,
```
graebm: Mention in the connect() docs that this stuff is copied now

source/client.c
```diff
@@ -491,12 +486,18 @@ int aws_mqtt_client_connection_connect(
...
     connection->port = port;
-    connection->tls_options = tls_options;
-    connection->socket_options = socket_options;
+    connection->socket_options = *socket_options;
     connection->state = AWS_MQTT_CLIENT_STATE_CONNECTING;
     connection->clean_session = clean_session;
     connection->keep_alive_time = keep_alive_time;
 
+    /* Cheat and set the tls_options host_name to our copy if they're the same */
+    if (tls_options) {
```
graebm: Since connect() can theoretically be called multiple times, should we zero out connection->tls_options if NULL was passed in?

graebm:
"instead of making other people own them" <-- would you describe this as "The Uber of Github PRs"

source/client.c
ColdenCullen: Yes!

include/aws/mqtt/client.h
ColdenCullen: Yes!

PR #58, opened 2019-03-20
Automatic ping and interupt notification fixes.

JonathanHenson:

### DIFF
--- a/include/aws/mqtt/client.h
+++ b/include/aws/mqtt/client.h
@@ -122,23 +122,22 @@ struct aws_mqtt_topic_subscription {
  * clean_session             True to discard all server session data and start fresh
  * keep_alive_time_secs      The keep alive value to place in the CONNECT PACKET, a PING will automatically
  *                           be sent at this interval as well. If you specify 0, defaults will be used
- *                           and a ping will be sent every 60s. If you're tempted to set this really low
- *                           to find out about connection drops quickly, this is going to consume a lot of cpu.
- *                           Instead, consider setting tcp keep alive to low timeout values and leaving the ping
- *                           at the max value allowed by the broker.
- * request_timeout_ms        Amount of time before a PING is considered timed out. Default is 3s.
- *                           Specify 0 for defaults.
+ *                           and a ping will be sent once per 60 minutes.
+ * ping_timeout_ms           Network connection is re-established if a ping response is not received within
+ *                           this amount of time (milliseconds). If you specify 0, a default value of 3 seconds is used.
+ *                           Alternatively, tcp keep-alive may be away to accomplish this in a more efficient
+ * (low-power) scenario, but keep-alive options may not work the same way on every platform and OS version.
  * on_connection_complete    The callback to fire when the connection attempt completes
  * user_data                  Passed to the userdata param of on_connection_complete
  */
 struct aws_mqtt_connection_options {
-    const struct aws_byte_cursor *host_name;
+    const struct aws_byte_cursor host_name;
     uint16_t port;
     const struct aws_socket_options *socket_options;
     const struct aws_tls_connection_options *tls_options;
-    const struct aws_byte_cursor *client_id;
+    const struct aws_byte_cursor client_id;
     uint16_t keep_alive_time_secs;
-    uint32_t request_timeout_ms;
+    uint32_t ping_timeout_ms;
     aws_mqtt_client_on_connection_complete_fn *on_connection_complete;
     void *user_data;
     bool clean_session;
--- a/include/aws/mqtt/client.h
+++ b/include/aws/mqtt/client.h
@@ -110,6 +110,40 @@ struct aws_mqtt_topic_subscription {
     void *on_publish_ud;
 };
 
+/**
+ * host_name                 The server name to connect to. This resource may be freed immediately on return.
+ * port                      The port on the server to connect to
+ * client_id                 The clientid to place in the CONNECT packet.
+ * socket_options            The socket options to pass to the aws_client_bootstrap functions.
+ *                           This is copied into the connection
+ * tls_options               TLS settings to use when opening a connection.
+ *                           This is copied into the connection
+ *                           Pass NULL to connect without TLS (NOT RECOMMENDED)
+ * clean_session             True to discard all server session data and start fresh
+ * keep_alive_time           The keep alive value to place in the CONNECT PACKET, a PING will automatically
+ *                           be sent at this interval as well. If you specify 0, defaults will be used
+ *                           and a ping will be sent every 60s. If you're tempted to set this really low
+ *                           to find out about connection drops quickly, this is going to consume a lot of cpu.
+ *                           Instead, consider setting tcp keep alive to low timeout values and leaving the ping
+ *                           at the max value allowed by the broker.
+ * request_timeout_ms        Amount of time before a PING is considered timed out. Default is 3s.
+ *                           Specify 0 for defaults.
+ * on_connection_complete    The callback to fire when the connection attempt completes
+ * user_data                  Passed to the userdata param of on_connection_complete
+ */
+struct aws_mqtt_connection_options {
+    const struct aws_byte_cursor *host_name;
+    uint16_t port;
+    const struct aws_socket_options *socket_options;
+    const struct aws_tls_connection_options *tls_options;
+    const struct aws_byte_cursor *client_id;
+    uint16_t keep_alive_time_secs;
+    uint32_t request_timeout_ms;
+    aws_mqtt_client_on_connection_complete_fn *on_connection_complete;
+    void *user_data;
+    bool clean_session;
+};
+
 AWS_EXTERN_C_BEGIN
 
 /**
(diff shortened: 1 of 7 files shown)


source/client_channel_handler.c
```diff
@@ -41,6 +41,27 @@ static int s_packet_handler_default(
...
+
+    if (status == AWS_TASK_STATUS_RUN_READY) {
+        struct aws_mqtt_client_connection *connection = arg;
+        aws_mqtt_client_connection_ping(connection);
+        aws_channel_task_init(&connection->ping_task, s_on_time_to_ping, connection);
+
+        uint64_t schedule_time = 0;
+        aws_channel_current_clock_time(connection->slot->channel, &schedule_time);
+        if (connection->keep_alive_time) {
+            schedule_time +=
```
justinboswell: Should we try to reduce the task time so that the ping is sent just under the timeout value, so that there's no chance of latency causing a hangup?

include/aws/mqtt/client.h
```diff
@@ -245,6 +249,7 @@ int aws_mqtt_client_connection_connect(
     const struct aws_byte_cursor *client_id,
     bool clean_session,
     uint16_t keep_alive_time,
+    uint32_t request_timeout_ms,
```
justinboswell: While we're messing with this, should we put these (clean_session, keep_alive_time, request_timeout_ms) into a ```mqtt_connection_options``` struct so the API doesn't change after this, should we need a new config value?

source/client_channel_handler.c
JonathanHenson: I subtracted the request timeout out of the to run time. The only downside here is if someone does something dumb and pings every second with a 3 second timeout, they'll be running pings all day.

JonathanHenson: but now that TCP keep alive can successfully pick up dropped connections within a second, we can just advise people to let this be set at the max value.

source/client.c
```diff
@@ -82,6 +84,9 @@ static void s_mqtt_client_init(
                 aws_event_loop_group_get_next_loop(connection->client->bootstrap->event_loop_group);
             aws_event_loop_schedule_task_future(
                 el, &connection->reconnect_task->task, connection->reconnect_timeouts.next_attempt);
+        } else {
+            connection->state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
+            MQTT_CLIENT_CALL_CALLBACK_ARGS(connection, on_connection_complete, error_code, 0, false);
```
graebm: nice catch, surprised we missed this before

include/aws/mqtt/client.h
```diff
@@ -228,7 +228,11 @@ int aws_mqtt_client_connection_set_connection_interruption_handlers(
  *                                          This is copied into the connection
  *                                          Pass NULL to connect without TLS (NOT RECOMMENDED)
  * \param[in] clean_session             True to discard all server session data and start fresh
- * \param[in] keep_alive_time           The keep alive value to place in the CONNECT PACKET
+ * \param[in] keep_alive_time           The keep alive value to place in the CONNECT PACKET, a PING will automatically
+ *                                       be sent at this interval as well. If you specify 0, defaults will be used
+ *                                       and a ping will be sent every 60s.
+ * \param[in] request_timeout_ms        Amount of time before a message is considered timed out. Default is 3s.
+ *                                      Specify 0 for defaults.
```
graebm: This sounds like it applies to normal PUB/SUB MQTT messages, which shouldn't have timeouts in this async API. If they're QoS 1, then the client should just keep them queued until a reconnect is finally established.

This is only supposed to apply to the keepalive ping, right?

source/client_channel_handler.c
graebm: How about we just keep using the one "keepalive" value, instead of introducing a second value. Here's what the MQTT spec says about keepalive on the server-side:
> If the Keep Alive value is non-zero and the Server does not receive a Control Packet from the Client within one and a half times the Keep Alive time period, it MUST disconnect the Network Connection to the Client as if the network had failed [MQTT-3.1.2-24].

We could follow that approach too. If we send a PINGREQ every 10 seconds, it should take 15 seconds without a PINGRESP before we kill the connection. It will confuse users that keepalive 10s means it takes 15s to timeout, but we could just document that. Or we could let user-passed-keepalive of 10s translate to pings every 6.66s, but that's not "literal mqtt keepalive" behavior. Choose the form of your confusor I guess

source/client.c
```diff
@@ -32,6 +32,8 @@
 #    pragma warning(disable : 4204)
 #endif
 
+static const uint64_t s_default_request_timeout_ns = 3000000000;
```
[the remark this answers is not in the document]
JonathanHenson: the iot service interprets 0 as a value of an hour or something like that.

Also, this value isn't for keep alive, it's how long we wait for a response before assuming the connection is dead.

include/aws/mqtt/client.h
JonathanHenson: i'll update the comment.

source/client_channel_handler.c
graebm: Actually, the spec says that ANY packet type is sufficient to keep the connection alive, not just pings. As an optimization would could only send pings when no other kind of packet has been sent in keepalive seconds

graebm: > It is the responsibility of the Client to ensure that the interval between Control Packets being sent does not exceed the Keep Alive value. In the absence of sending any other Control Packets, the Client MUST send a PINGREQ Packet

JonathanHenson: I don't think this one is relevant anymore, closing.

include/aws/mqtt/client.h
```diff
@@ -110,6 +110,40 @@ struct aws_mqtt_topic_subscription {
...
+ *                           to find out about connection drops quickly, this is going to consume a lot of cpu.
+ *                           Instead, consider setting tcp keep alive to low timeout values and leaving the ping
+ *                           at the max value allowed by the broker.
+ * request_timeout_ms        Amount of time before a PING is considered timed out. Default is 3s.
+ *                           Specify 0 for defaults.
+ * on_connection_complete    The callback to fire when the connection attempt completes
+ * user_data                  Passed to the userdata param of on_connection_complete
+ */
+struct aws_mqtt_connection_options {
+    const struct aws_byte_cursor *host_name;
```
bretambrose: I have a lot of options structs/parameters that use aws_string pointers (with non-owner-transfer semantics).  Would you advise me to follow suit and use byte cursors for those?

include/aws/mqtt/client.h
```diff
@@ -110,6 +110,40 @@ struct aws_mqtt_topic_subscription {
...
+ *                           This is copied into the connection
+ *                           Pass NULL to connect without TLS (NOT RECOMMENDED)
+ * clean_session             True to discard all server session data and start fresh
+ * keep_alive_time           The keep alive value to place in the CONNECT PACKET, a PING will automatically
+ *                           be sent at this interval as well. If you specify 0, defaults will be used
+ *                           and a ping will be sent every 60s. If you're tempted to set this really low
+ *                           to find out about connection drops quickly, this is going to consume a lot of cpu.
+ *                           Instead, consider setting tcp keep alive to low timeout values and leaving the ping
+ *                           at the max value allowed by the broker.
+ * request_timeout_ms        Amount of time before a PING is considered timed out. Default is 3s.
```
bretambrose: If this only applies to pings maybe update name to "ping_timeout_ms"  If it applies to more than pings than maybe adjust comment.  Related, are pings the only request/response component of mqtt?

source/client.c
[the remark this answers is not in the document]
graebm: Sounds like it's disconnecting "disconnecting a Client that it determines to be inactive or non-responsive at any time, regardless of the Keep Alive value provided by that Client."

source/client.c
```diff
@@ -488,16 +486,23 @@ int aws_mqtt_client_connection_connect(
...
+    } else {
+        connection->request_timeout_ns = aws_timestamp_convert(
+            (uint64_t)connection_options->request_timeout_ms, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_NANOS, NULL);
+    }
 
     /* Cheat and set the tls_options host_name to our copy if they're the same */
-    if (tls_options) {
-        connection->tls_options = *tls_options;
+    if (connection_options->tls_options) {
+        connection->tls_options = *connection_options->tls_options;
```
bretambrose: Doesn't this need to use aws_tls_connection_options_copy?

source/client_channel_handler.c
```diff
@@ -41,6 +42,32 @@ static int s_packet_handler_default(
     return aws_raise_error(AWS_ERROR_MQTT_INVALID_PACKET_TYPE);
 }
 
+static void s_on_time_to_ping(struct aws_channel_task *channel_task, void *arg, enum aws_task_status status);
+static void s_schedule_ping(struct aws_mqtt_client_connection *connection) {
+    aws_channel_task_init(&connection->ping_task, s_on_time_to_ping, connection);
+
+    uint64_t schedule_time = 0;
+    aws_channel_current_clock_time(connection->slot->channel, &schedule_time);
+    if (connection->keep_alive_time) {
```
bretambrose: Minor, but request_timeout gets initialized to the default value if its options value comes in as zero.  Seems reasonable to do the same thing with this setting so that the time calculation becomes unconditional.

include/aws/mqtt/client.h
JonathanHenson: I dunno, this isn't really a change from the existing code, it just moved stuff around.

these are always copied, so they should be cursors.

include/aws/mqtt/client.h
JonathanHenson: I don't know all the ways they apply because I only touched PING, but in general, there's probably a timeout (at least in the future) on suback and puback.

source/client.c
JonathanHenson: no, it gets copied by the bootstrap.

source/client_channel_handler.c
JonathanHenson: good point.

include/aws/mqtt/client.h
```diff
@@ -110,6 +110,40 @@ struct aws_mqtt_topic_subscription {
...
+ *                           Pass NULL to connect without TLS (NOT RECOMMENDED)
+ * clean_session             True to discard all server session data and start fresh
+ * keep_alive_time_secs      The keep alive value to place in the CONNECT PACKET, a PING will automatically
+ *                           be sent at this interval as well. If you specify 0, defaults will be used
+ *                           and a ping will be sent every 60s. If you're tempted to set this really low
+ *                           to find out about connection drops quickly, this is going to consume a lot of cpu.
+ *                           Instead, consider setting tcp keep alive to low timeout values and leaving the ping
+ *                           at the max value allowed by the broker.
+ * request_timeout_ms        Amount of time before a PING is considered timed out. Default is 3s.
+ *                           Specify 0 for defaults.
```
graebm: Be super clear that this results in a reconnect attempt, like
> ping_timeout_ms:       Network connection is re-established if a ping response is not received within this amount of time (milliseconds). If you specify 0, a default value of 3 seconds is used.

note I changed the variable name too

source/client_channel_handler.c
JonathanHenson: nevermind, this doesn't actually apply, because keep alive has meaning on the connect, and reconnect calls so I don't want to redefine it.

include/aws/mqtt/client.h
```diff
@@ -110,6 +110,40 @@ struct aws_mqtt_topic_subscription {
...
+ *                           This is copied into the connection
+ * tls_options               TLS settings to use when opening a connection.
+ *                           This is copied into the connection
+ *                           Pass NULL to connect without TLS (NOT RECOMMENDED)
+ * clean_session             True to discard all server session data and start fresh
+ * keep_alive_time_secs      The keep alive value to place in the CONNECT PACKET, a PING will automatically
+ *                           be sent at this interval as well. If you specify 0, defaults will be used
+ *                           and a ping will be sent every 60s. If you're tempted to set this really low
+ *                           to find out about connection drops quickly, this is going to consume a lot of cpu.
+ *                           Instead, consider setting tcp keep alive to low timeout values and leaving the ping
```
graebm: after our conversation, I think it's worth keeping the comments that tcp-timeout might be a lower-power way to accomplish this, but might not work on every platform

include/aws/mqtt/client.h
```diff
@@ -110,6 +110,40 @@ struct aws_mqtt_topic_subscription {
...
+ *                           at the max value allowed by the broker.
+ * request_timeout_ms        Amount of time before a PING is considered timed out. Default is 3s.
+ *                           Specify 0 for defaults.
+ * on_connection_complete    The callback to fire when the connection attempt completes
+ * user_data                  Passed to the userdata param of on_connection_complete
+ */
+struct aws_mqtt_connection_options {
+    const struct aws_byte_cursor *host_name;
```
graebm: byte_cursors on the options struct probably shouldn't be pointers

justinboswell: Good catch, I totally missed that :(

PR #64, opened 2019-04-09
Fix leaking TLS connection options and reconnect_task

ColdenCullen:

### DIFF
--- a/source/client_channel_handler.c
+++ b/source/client_channel_handler.c
@@ -690,7 +690,8 @@ static void s_mqtt_disconnect_task(struct aws_channel_task *channel_task, void *
 void mqtt_disconnect_impl(struct aws_mqtt_client_connection *connection, int error_code) {
 
     if (connection->slot) {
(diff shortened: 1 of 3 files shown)


source/client_channel_handler.c
```diff
@@ -663,6 +663,12 @@ void mqtt_disconnect_impl(struct aws_mqtt_client_connection *connection, int err
     /* If there is an outstanding reconnect task, cancel it */
     if (connection->state == AWS_MQTT_CLIENT_STATE_DISCONNECTING && connection->reconnect_task) {
         aws_atomic_store_ptr(&connection->reconnect_task->connection_ptr, NULL);
+
+        /* If the reconnect_task isn't scheduled, free it */
+        if (connection->reconnect_task && !connection->reconnect_task->task.timestamp) {
```
justinboswell: Is there any chance that the timestamp could be written from another thread at this point? If so, this probably isn't protected enough.

ColdenCullen: moving this to a task I guess

PR #68, opened 2019-04-24
Add logging to MQTT

ColdenCullen:

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -45,6 +45,8 @@ int aws_mqtt_client_init(
     struct aws_allocator *allocator,
     struct aws_client_bootstrap *bootstrap) {
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Initalizing MQTT client");
+
     AWS_ZERO_STRUCT(*client);
     client->allocator = allocator;
     client->bootstrap = bootstrap;
@@ -70,6 +72,8 @@ static void s_mqtt_client_shutdown(
     (void)bootstrap;
     (void)channel;
 
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Channel has been shutdown with error code %d", error_code);
+
     struct aws_mqtt_client_connection *connection = user_data;
 
     /* Always clear slot, as that's what's been shutdown */
@@ -88,6 +92,8 @@ static void s_mqtt_client_shutdown(
         /* If reconnect attempt failed, schedule the next attempt */
         struct aws_event_loop *el = aws_event_loop_group_get_next_loop(connection->client->bootstrap->event_loop_group);
 
+        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Reconnect failed, retrying");
+
         aws_event_loop_schedule_task_future(
             el, &connection->reconnect_task->task, connection->reconnect_timeouts.next_attempt);
 
@@ -95,12 +101,16 @@ static void s_mqtt_client_shutdown(
 
         connection->state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
 
+        AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Disconnect completed, clearing request queue and calling callback");
+
         /* Successfully shutdown, so clear the outstanding requests */
         aws_hash_table_clear(&connection->outstanding_requests.table);
 
         MQTT_CLIENT_CALL_CALLBACK(connection, on_disconnect);
 
     } else if (connection->state == AWS_MQTT_CLIENT_STATE_CONNECTING) {
+
+        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Initial connection attempt failed, calling callback");
 
         connection->state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
         MQTT_CLIENT_CALL_CALLBACK_ARGS(connection, on_connection_complete, error_code, 0, false);
@@ -114,6 +124,8 @@ static void s_mqtt_client_shutdown(
 
         if (connection->state == AWS_MQTT_CLIENT_STATE_CONNECTED) {
 
+            AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Connection lost, calling callback and attempting reconnect");
+
             connection->state = AWS_MQTT_CLIENT_STATE_RECONNECTING;
             MQTT_CLIENT_CALL_CALLBACK_ARGS(connection, on_interrupted, error_code);
         }
@@ -125,6 +137,10 @@ static void s_mqtt_client_shutdown(
 
         /* This will only be true if the user called disconnect from the on_interrupted callback */
         if (connection->state == AWS_MQTT_CLIENT_STATE_DISCONNECTING) {
+
+            AWS_LOGF_TRACE(
+                AWS_LS_MQTT_CLIENT, "Caller requested disconnect from on_interrupted callback, aborting reconnect");
+
             connection->state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
             MQTT_CLIENT_CALL_CALLBACK(connection, on_disconnect);
 
@@ -160,6 +176,8 @@ static void s_mqtt_client_init(
         return;
     }
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Connection successfully opened, sending CONNECT packet");
+
     /* Reset the current timeout timer */
     connection->reconnect_timeouts.current = connection->reconnect_timeouts.min;
 
@@ -167,6 +185,7 @@ static void s_mqtt_client_init(
     connection->slot = aws_channel_slot_new(channel);
 
     if (!connection->slot) {
+        AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "Failed to create new slot, something has gone horribly wrong");
         aws_channel_shutdown(channel, aws_last_error());
         return;
     }
@@ -239,6 +258,8 @@ static void s_attempt_reconect(struct aws_task *task, void *userdata, enum aws_t
     if (status == AWS_TASK_STATUS_RUN_READY && connection) {
         /* If the task is not cancelled and a connection has not succeeded, attempt reconnect */
 
+        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Attempting reconnect");
+
         aws_high_res_clock_get_ticks(&connection->reconnect_timeouts.next_attempt);
         connection->reconnect_timeouts.next_attempt += aws_timestamp_convert(
             connection->reconnect_timeouts.current, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
@@ -290,10 +311,13 @@ struct aws_mqtt_client_connection *aws_mqtt_client_connection_new(struct aws_mqt
 
     assert(client);
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Creating new connection");
+
     struct aws_mqtt_client_connection *connection =
         aws_mem_acquire(client->allocator, sizeof(struct aws_mqtt_client_connection));
 
     if (!connection) {
+        AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "Failed to allocate connection object");
         return NULL;
     }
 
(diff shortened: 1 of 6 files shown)


source/client.c
```diff
@@ -289,10 +311,13 @@ struct aws_mqtt_client_connection *aws_mqtt_client_connection_new(struct aws_mqt
 
     assert(client);
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Creating new connection");
```
justinboswell: Probably worth adding some contextual info here, like endpoint/port at least?

Also, most of these need to have the id=%p so that multiple connections can be differentiated

source/client.c
```diff
@@ -413,12 +442,16 @@ int aws_mqtt_client_connection_set_will(
     bool retain,
     const struct aws_byte_cursor *payload) {
 
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Setting last will and testament");
```
justinboswell: topic?

source/client.c
```diff
@@ -989,6 +1056,8 @@ uint16_t aws_mqtt_client_connection_subscribe(
     uint16_t packet_id =
         mqtt_create_request(task_arg->connection, &s_subscribe_send, task_arg, &s_subscribe_single_complete, task_arg);
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Starting subscribe %d", packet_id);
```
justinboswell: topic?

source/client.c
```diff
@@ -867,6 +930,8 @@ uint16_t aws_mqtt_client_connection_subscribe_multiple(
     uint16_t packet_id =
         mqtt_create_request(task_arg->connection, &s_subscribe_send, task_arg, &s_subscribe_complete, task_arg);
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Starting subscribe %d", packet_id);
```
justinboswell: topic?

source/mqtt.c
```diff
@@ -147,10 +149,28 @@ void aws_mqtt_load_error_strings() {
...
+
+        static struct aws_log_subject_info_list s_logging_subjects_list = {
+            .subject_list = s_logging_subjects,
+            .count = AWS_ARRAY_SIZE(s_logging_subjects),
+        };
+        aws_register_log_subject_info_list(&s_logging_subjects_list);
     }
 }
+
+void aws_mqtt_load_error_strings(void) {
```
justinboswell: This should be the other way around, and we should fix the downstream libs

source/topic_tree.c
```diff
@@ -62,30 +64,27 @@ static struct topic_tree_action *s_topic_tree_action_create(struct aws_array_lis
...
-        goto handle_error;
+        AWS_LOGF_ERROR(AWS_LS_MQTT_TOPIC_TREE, "Failed to insert action into transaction, array_list_push_back failed");
+        goto push_back_failed;
     }
-    was_pushed = true;
 
     if (aws_array_list_get_at_ptr(transaction, (void **)&action, aws_array_list_length(transaction) - 1)) {
 
-        goto handle_error;
+        AWS_LOGF_ERROR(AWS_LS_MQTT_TOPIC_TREE, "Failed to retreive most recent action from transaction");
```
justinboswell: retrieve

source/topic_tree.c
```diff
@@ -282,6 +299,8 @@ static void s_topic_tree_action_commit(struct topic_tree_action *action, struct
 
                 /* "unsubscribe" current. */
                 if (current->cleanup && current->userdata) {
+                    AWS_LOGF_TRACE(AWS_LS_MQTT_TOPIC_TREE, "Cleaning up node's userdata");
```
justinboswell: id=%p and current would be helpful, perhaps?

source/client.c
```diff
@@ -69,6 +72,8 @@ static void s_mqtt_client_shutdown(
     (void)bootstrap;
     (void)channel;
 
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Channel has been shutdown with error code %d", error_code);
```
bretambrose: I'd move this forward a line and add the connection state to the log message since so much logic is conditionally executed on it.  Might be able to remove some of the below messages that are essentially just state indicators too if you switched this to DEBUG.

source/client.c
```diff
@@ -159,13 +176,16 @@ static void s_mqtt_client_init(
...
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Connection successfully opened, sending CONNECT packet");
+
     /* Reset the current timeout timer */
     connection->reconnect_timeouts.current = connection->reconnect_timeouts.min;
 
     /* Create the slot and handler */
     connection->slot = aws_channel_slot_new(channel);
 
     if (!connection->slot) {
+        AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "Failed to create new slot, something has gone horribly wrong");
```
bretambrose: is the handle_error: block below worth adding a log message in?

source/client.c
ColdenCullen: Will add id, but endpoint:port doesn't come until connect()

source/client.c
```diff
@@ -238,6 +258,8 @@ static void s_attempt_reconect(struct aws_task *task, void *userdata, enum aws_t
     if (status == AWS_TASK_STATUS_RUN_READY && connection) {
         /* If the task is not cancelled and a connection has not succeeded, attempt reconnect */
 
+        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Attempting reconnect");
```
bretambrose: It may be worth logging the no-reconnect case and why (max exceeded, canceled, etc...)

source/client.c
```diff
@@ -469,6 +508,8 @@ int aws_mqtt_client_connection_set_reconnect_timeout(
 
     assert(connection);
 
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Setting reconnect min and max timeout");
```
graebm: worth printing values?

source/client.c
```diff
@@ -289,10 +311,13 @@ struct aws_mqtt_client_connection *aws_mqtt_client_connection_new(struct aws_mqt
 
     assert(client);
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Creating new connection");
+
     struct aws_mqtt_client_connection *connection =
         aws_mem_acquire(client->allocator, sizeof(struct aws_mqtt_client_connection));
 
     if (!connection) {
+        AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "Failed to allocate connection object");
```
bretambrose: Usually don't bother with logging allocation failures since an allocation failure means the log command itself will most likely fail

source/client.c
```diff
@@ -308,18 +333,21 @@ struct aws_mqtt_client_connection *aws_mqtt_client_connection_new(struct aws_mqt
 
     if (aws_mutex_init(&connection->pending_requests.mutex)) {
 
-        goto handle_error;
+        AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "Failed to initialize pending_requests mutex");
+        goto failed_init_pending_requests_mutex;
```
bretambrose: Given the ZERO_STRUCT call here, it seems like this could be cleaned up to have a single on_error that destroys the connection, which would simplify the logic a lot.  Your destroy might need to be made a bit more paranoid in spots to make sure it handles the partial construction cases.

source/client.c
ColdenCullen: Can't without generating a big string as this is the multi-topic subscribe

source/mqtt.c
ColdenCullen: What? I followed the HTTP convention, is that wrong?

source/client.c
graebm: I think it would be very handy for users (and us helping people on Github) if the topic of each subscription were logged. Something like:
"id=%p packet=%d i=%d: Subscribing to topic: %s"

we'd need to log it at DEBUG or TRACE level though, not INFO, since it's user-generated data

unless this is already being logged at DEBUG level by the topic tree...

source/client.c
```diff
@@ -798,6 +859,8 @@ static void s_subscribe_complete(
 
     struct subscribe_task_arg *task_arg = userdata;
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Subscribe %d completed", packet_id);
```
bretambrose: This seems like an int size mismatch but somehow the gcc compiler check isn't complaining?

bretambrose: Also, maybe log error vs. no-error state (several spots in all these _complete callbacks)

source/client.c
```diff
@@ -907,6 +972,8 @@ static void s_subscribe_single_complete(
...
```
graebm: Log whether it was accepted or rejected
a rejected subscription might be worthy of a higher log level like WARNING?

source/client.c
```diff
@@ -1101,6 +1176,8 @@ static void s_unsubscribe_complete(
 
     struct unsubscribe_task_arg *task_arg = userdata;
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Unsubscribe %d complete", packet_id);
```
bretambrose: This is a strange if-statement

source/client.c
ColdenCullen: I like the more specific debug messages, and this general trace, but I'll add the connection id.

source/client.c
ColdenCullen: Nah, I'll just add logs before each error case

source/client.c
ColdenCullen: That makes sense

source/client.c
ColdenCullen: There isn't a no-reconnect case

source/mqtt.c
justinboswell: load_error_strings shouldn't be the entry point, library_init should. If HTTP is doing that, it's backwards and we should fix it. @graebm ?

source/client.c
ColdenCullen: Can fix, I'll use `PRIu16`

source/client.c
```diff
@@ -258,12 +259,15 @@ static void s_attempt_reconect(struct aws_task *task, void *userdata, enum aws_t
     if (status == AWS_TASK_STATUS_RUN_READY && connection) {
         /* If the task is not cancelled and a connection has not succeeded, attempt reconnect */
 
-        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Attempting reconnect");
-
         aws_high_res_clock_get_ticks(&connection->reconnect_timeouts.next_attempt);
         connection->reconnect_timeouts.next_attempt += aws_timestamp_convert(
             connection->reconnect_timeouts.current, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
 
+        AWS_LOGF_TRACE(
```
justinboswell: Needs id

source/client.c
```diff
@@ -87,20 +92,26 @@ static void s_mqtt_client_shutdown(
         /* If reconnect attempt failed, schedule the next attempt */
         struct aws_event_loop *el = aws_event_loop_group_get_next_loop(connection->client->bootstrap->event_loop_group);
 
+        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "Reconnect failed, retrying");
```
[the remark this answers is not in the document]
ColdenCullen: I disagree, I don't think people need to know about every time a reconnect fails, only when the connection goes offline, an when it goes back online again.

source/mqtt.c
ColdenCullen: Nuked load_error_strings per mike

graebm:
Would you mind if I harassed you to paste what the logs looked like at TRACE/DEBUG/INFO levels when running something simple like a pub/sub sample? connect, sub, pub, receive, disconnect

source/client.c
```diff
@@ -57,6 +61,8 @@ int aws_mqtt_client_init(
 
 void aws_mqtt_client_clean_up(struct aws_mqtt_client *client) {
 
+    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "client=%p Cleaning up MQTT client", (void *)client);
```
graebm: The first two log messages have "client=%p" and later ones use "id=%p"
Should probably stick with "id=%p". The client already has its own subject string, and "id" is we've been doing in other systems

source/client.c
```diff
@@ -71,6 +77,9 @@ static void s_mqtt_client_shutdown(
 
     struct aws_mqtt_client_connection *connection = user_data;
 
+    AWS_LOGF_TRACE(
+        AWS_LS_MQTT_CLIENT, "id=%p Channel has been shutdown with error code %d", (void *)connection, error_code);
```
graebm: Other systems put a colon after the identifiers like "id=%p: Content of message"

source/client.c
ColdenCullen: These are two different things, id= is used when referring to a connection, not the client.

PR #78, opened 2019-06-25
Addressed timeout timing bug with mqtt ping timeouts.

JonathanHenson:

### DIFF
--- a/include/aws/mqtt/client.h
+++ b/include/aws/mqtt/client.h
@@ -153,12 +153,11 @@ struct aws_mqtt_topic_subscription {
  * clean_session             True to discard all server session data and start fresh
  * keep_alive_time_secs      The keep alive value to place in the CONNECT PACKET, a PING will automatically
  *                           be sent at this interval as well. If you specify 0, defaults will be used
- *                           and a ping will be sent once per 60 minutes.
- * ping_timeout_ms           Network connection is re-established if a ping response is not received within
- *                           this amount of time (milliseconds). If you specify 0, a default value of 3 seconds is used.
- *                           Alternatively, tcp keep-alive may be away to accomplish this in a more efficient
- *                           (low-power) scenario, but keep-alive options may not work the same way on every platform
- *                           and OS version.
+ *                           and a ping will be sent once per 60 minutes. This value must be greater than
+ * ping_timeout_ms ping_timeout_ms           Network connection is re-established if a ping response is not received
+ * within this amount of time (milliseconds). If you specify 0, a default value of 3 seconds is used. Alternatively, tcp
+ * keep-alive may be away to accomplish this in a more efficient (low-power) scenario, but keep-alive options may not
+ * work the same way on every platform and OS version. This value must be less than keep_alive_time_secs.
  * on_connection_complete    The callback to fire when the connection attempt completes
  * user_data                  Passed to the userdata param of on_connection_complete
  */
--- a/include/aws/mqtt/client.h
+++ b/include/aws/mqtt/client.h
@@ -429,18 +429,6 @@ uint16_t aws_mqtt_client_connection_publish(
     aws_mqtt_op_complete_fn *on_complete,
     void *userdata);
 
-/**
- * Sends a PINGREQ packet to the server to keep the connection alive.
- * If a PINGRESP is not received within a reasonable period of time, the connection will be closed.
- *
- * \params[in] connection   The connection to ping on
- *
- * \returns AWS_OP_SUCCESS if the connection is open and the PINGREQ is sent or queued to send,
- *              otherwise AWS_OP_ERR and aws_last_error() is set.
- */
-AWS_MQTT_API
-int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connection);
-
 AWS_EXTERN_C_END
 
 #endif /* AWS_MQTT_CLIENT_H */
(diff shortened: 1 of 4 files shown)


source/client.c
```diff
@@ -851,6 +851,19 @@ int aws_mqtt_client_connection_connect(
             (uint64_t)connection_options->ping_timeout_ms, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_NANOS, NULL);
     }
 
+    /* Keep alive time should always be greater than the timeouts. */
+    if (AWS_UNLIKELY(
```
justinboswell: Couldn't this be a warning, and we just clamp the value for you?

source/client.c
```diff
@@ -851,6 +851,19 @@ int aws_mqtt_client_connection_connect(
...
+            connection->keep_alive_time_secs * (uint64_t)AWS_TIMESTAMP_NANOS <= connection->request_timeout_ns)) {
+        AWS_LOGF_FATAL(
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Illegal configuration, Connection keep alive %" PRIu16
+            "s must be greater than the request timeouts %" PRIu64 "ns.",
+            (void *)connection,
+            connection->keep_alive_time_secs,
+            connection->request_timeout_ns);
+        AWS_FATAL_ASSERT(0);
```
justinboswell: If we keep this, just assert the condition above, so that the assert is helpful.

source/client.c
```diff
@@ -1735,22 +1748,17 @@ static enum aws_mqtt_client_request_state s_pingreq_send(uint16_t message_id, bo
...
     }
 
     /* Check that a pingresp has been received since pingreq was sent */
-
-    uint64_t current_time = 0;
-    aws_channel_current_clock_time(connection->slot->channel, &current_time);
-
-    if (current_time - connection->last_pingresp_timestamp > connection->request_timeout_ns) {
+    if (connection->waiting_on_ping_response) {
+        connection->waiting_on_ping_response = false;
```
justinboswell: What if multiple pings are in flight? Should this maybe tick up when we send, and down when we receive?

include/aws/mqtt/client.h
```diff
@@ -153,12 +153,11 @@ struct aws_mqtt_topic_subscription {
...
  * keep_alive_time_secs      The keep alive value to place in the CONNECT PACKET, a PING will automatically
  *                           be sent at this interval as well. If you specify 0, defaults will be used
- *                           and a ping will be sent once per 60 minutes.
- * ping_timeout_ms           Network connection is re-established if a ping response is not received within
- *                           this amount of time (milliseconds). If you specify 0, a default value of 3 seconds is used.
- *                           Alternatively, tcp keep-alive may be away to accomplish this in a more efficient
- *                           (low-power) scenario, but keep-alive options may not work the same way on every platform
- *                           and OS version.
+ *                           and a ping will be sent once per 60 minutes. This value must be greater than
+ * ping_timeout_ms ping_timeout_ms           Network connection is re-established if a ping response is not received
```
bretambrose: Seems like some formatting got mangled here

include/aws/mqtt/client.h
```diff
@@ -153,12 +153,11 @@ struct aws_mqtt_topic_subscription {
...
+ * within this amount of time (milliseconds). If you specify 0, a default value of 3 seconds is used. Alternatively, tcp
+ * keep-alive may be away to accomplish this in a more efficient (low-power) scenario, but keep-alive options may not
```
bretambrose: might as well correct "away" -> "a way" while here

source/client.c
```diff
@@ -851,6 +851,19 @@ int aws_mqtt_client_connection_connect(
             (uint64_t)connection_options->ping_timeout_ms, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_NANOS, NULL);
     }
 
+    /* Keep alive time should always be greater than the timeouts. */
+    if (AWS_UNLIKELY(
+            connection->keep_alive_time_secs * (uint64_t)AWS_TIMESTAMP_NANOS <= connection->request_timeout_ns)) {
+        AWS_LOGF_FATAL(
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Illegal configuration, Connection keep alive %" PRIu16
+            "s must be greater than the request timeouts %" PRIu64 "ns.",
```
bretambrose: It's annoying, but I'd make sure both time values were in the same unit here, to make interpretation easier and less likely to be wrong.

source/client.c
JonathanHenson: not possible now since PING is being removed from the public API

source/client.c
JonathanHenson: That's going to be a hidden behavior that we can't actually get right 100% of the time. We already have default values, if someone sets them, and they're wrong, I think we should let them know immediately.

justinboswell: Ok, fair enough, sounds good.

include/aws/mqtt/private/client_impl.h
```diff
@@ -201,4 +201,16 @@ AWS_MQTT_API void mqtt_request_complete(
...
+ * Sends a PINGREQ packet to the server to keep the connection alive. This is not exported and should not ever
+ * be called directly. This function is driven by the timeout values passed to aws_mqtt_client_connect().
+ * If a PINGRESP is not received within a reasonable period of time, the connection will be closed.
+ *
+ * \params[in] connection   The connection to ping on
+ *
+ * \returns AWS_OP_SUCCESS if the connection is open and the PINGREQ is sent or queued to send,
+ *              otherwise AWS_OP_ERR and aws_last_error() is set.
+ */
+int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connection);
```
justinboswell: Remove the aws_ since it's private now?

PR #87, opened 2019-10-02
Reconnect Resubscribe

ColdenCullen:
This PR implements topic tree traversal, as well as automatically resubscribing to topics on connection resume.


source/client.c
```diff
@@ -1788,3 +1779,150 @@ int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connectio
...
+        error_code);
+
+    aws_array_list_clean_up(&task_arg->topics);
+    aws_mqtt_packet_subscribe_clean_up(&task_arg->subscribe);
+    aws_mem_release(task_arg->connection->allocator, task_arg);
+}
+
+void mqtt_resubscribe_existing_topics(struct aws_mqtt_client_connection *connection) {
+    size_t sub_count = aws_mqtt_topic_tree_get_sub_count(&connection->subscriptions);
+    static const size_t sub_size = sizeof(struct aws_mqtt_topic_subscription);
```
graebm: is there any limit to the number of topics that can be in one SUBSCRIBE packet?

Also: HEY! our multisub support finally actually came in handy

source/client.c
```diff
@@ -1788,3 +1779,150 @@ int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connectio
...
+        return;
+    }
+
+    AWS_ZERO_STRUCT(*task_arg);
+    task_arg->connection = connection;
+    aws_array_list_init_static(&task_arg->topics, buffer, sub_count, sub_size);
+    aws_mqtt_topic_tree_iterate(&connection->subscriptions, s_reconnect_resub_iterator, task_arg);
+
+    uint16_t packet_id =
+        mqtt_create_request(task_arg->connection, &s_resubscribe_send, task_arg, &s_resubscribe_complete, task_arg);
```
graebm: not checking for failure

source/client.c
```diff
@@ -1788,3 +1779,150 @@ int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connectio
...
+    if (!task_arg) {
+        AWS_LOGF_FATAL(
+            AWS_LS_MQTT_CLIENT, "id=%p: failed to allocate storage for resubscribe arguments", (void *)connection);
+        return;
+    }
+
+    AWS_ZERO_STRUCT(*task_arg);
+    task_arg->connection = connection;
+    aws_array_list_init_static(&task_arg->topics, buffer, sub_count, sub_size);
+    aws_mqtt_topic_tree_iterate(&connection->subscriptions, s_reconnect_resub_iterator, task_arg);
```
graebm: is there anything that might go wrong in the iterator function, that we'd want to check for after iterating everything?

source/client.c
```diff
@@ -1788,3 +1779,150 @@ int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connectio
...
+        packet_id,
+        error_code);
+
+    aws_array_list_clean_up(&task_arg->topics);
+    aws_mqtt_packet_subscribe_clean_up(&task_arg->subscribe);
+    aws_mem_release(task_arg->connection->allocator, task_arg);
+}
+
+void mqtt_resubscribe_existing_topics(struct aws_mqtt_client_connection *connection) {
+    size_t sub_count = aws_mqtt_topic_tree_get_sub_count(&connection->subscriptions);
```
graebm: do we need to hold any locks to be sure the tree isn't modified during `mqtt_resubscribe_existing_topics()`

source/client.c
```diff
@@ -1788,3 +1779,150 @@ int aws_mqtt_client_connection_ping(struct aws_mqtt_client_connection *connectio
...
+
+static bool s_reconnect_resub_iterator(const struct aws_byte_cursor *topic, enum aws_mqtt_qos qos, void *user_data) {
+    struct resubscribe_task_arg *task_arg = user_data;
+
+    struct aws_mqtt_topic_subscription sub;
+    AWS_ZERO_STRUCT(sub);
+    sub.topic = *topic;
+    sub.qos = qos;
+
+    aws_array_list_push_back(&task_arg->topics, &sub);
```
graebm: What happens if the user's original subscribe hasn't gotten its suback callback yet? Is that subscribe command in a queue somewhere? Will issuing this giant resubscribe command somehow interfere with them receiving their suback calback? Are there complications that might arise when we haven't sent the initial SUBSCRIBE vs when we lost the connection before the SUBACK came back?

just peppering you with every scenario I can imagine

source/client.c
ColdenCullen: No, all tree modifications happen on the event loop thread (including resubscribe).

source/client.c
ColdenCullen: Nope! It's just pushing into an array list of a known good size.

source/client.c
ColdenCullen: That's a great question, I don't think so. I think the total length just needs to be less than the max packet size (256MB).

graebm: still worth checking?

graebm: we should detect the exception **somewhere**. If the encoder detected the packet was too big and dealt with it, that would work too. If the complexity of dealing with these errors is too extreme, and AWS_FATAL_ASSERT() would be an acceptable placeholder for the edge case

source/client.c
```diff
@@ -1923,6 +1923,15 @@ void mqtt_resubscribe_existing_topics(struct aws_mqtt_client_connection *connect
     uint16_t packet_id =
         mqtt_create_request(task_arg->connection, &s_resubscribe_send, task_arg, &s_resubscribe_complete, task_arg);
 
+    if (packet_id == 0) {
+        AWS_LOGF_DEBUG(
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Failed to send multi-topic resubscribe with error %s",
+            (void *)connection,
+            aws_error_name(aws_last_error()));
+        return;
```
graebm: Other than logging it out, we're not dealing with this or reporting it to the user in a way they can react to this now-not-super-valid connection

include/aws/mqtt/client.h
```diff
@@ -406,6 +406,21 @@ uint16_t aws_mqtt_client_connection_subscribe(
...
+ * Resubscribe to all topics currently subscribed to. This is to help when resuming a connection with a clean session.
+ *
+ * \param[in] connection    The connection to subscribe on
+ * \param[in] on_suback     Called when a SUBACK has been received from the server and the subscription is complete
+ * \param[in] on_suback_ud  Passed to on_suback
+ *
+ * \returns The packet id of the subscribe packet if successfully sent, otherwise 0 (and aws_last_error() will be set).
+ */
+AWS_MQTT_API
+uint16_t aws_mqtt_resubscribe_existing_topics(
```
graebm: Mention this in the comments/docs of `aws_mqtt_client_connection_set_connection_interruption_handlers` and aws_mqtt_client_on_connection_resumed_fn`. Something like

> If session_present is false, you probably want to call `aws_mqtt_resubscribe_existing_topics`.

source/client.c
```diff
@@ -1443,6 +1434,138 @@ uint16_t aws_mqtt_client_connection_subscribe(
...
+
+    struct subscribe_task_arg *task_arg = NULL;
+    void *buffer = NULL;
+    aws_mem_acquire_many(
+        connection->allocator, 2, &task_arg, sizeof(struct subscribe_task_arg), &buffer, sub_count * sub_size);
+
+    if (!task_arg) {
+        AWS_LOGF_FATAL(
+            AWS_LS_MQTT_CLIENT, "id=%p: failed to allocate storage for resubscribe arguments", (void *)connection);
+        return 0;
```
graebm: This isn't fatal anymore

PR #90, opened 2019-10-07
Disambiguate a failed resubscribe() from an unnecessary resubscribe()

graebm:
Mqtt client functions return packet_id instead of error code, with 0 indicating a failure. Resubscribe was returning 0 if there was no need to send a packet, but there was no way to know whether or not this was a real error.

Therefore, add a specific error code to cover this edge case. This is weird, since it's a harmless error, but it makes sense since to make this an error-code since there's also a completion callback that will never fire.

In aws-crt-python I plan to do this differently. The resubscribe fn returns `(packet_id, Future)`. Rather than throw, I plan to return `None` for a packet_id and set the `Future` to an empty list of topics.

Is it too weird to make opposite choices in 2 different languages? C can't return `None` like python can, and setting up a callback in C is way more work than processing a `Future` in python.


source/client.c
```diff
@@ -1527,6 +1527,12 @@ uint16_t aws_mqtt_resubscribe_existing_topics(
     static const size_t sub_size = sizeof(struct aws_mqtt_topic_subscription);
 
     if (sub_count == 0) {
+        aws_raise_error(AWS_ERROR_MQTT_NO_TOPICS_FOR_RESUBSCRIBE);
+        AWS_LOGF_WARN(
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Resubscribe is unnecessary, error %s.",
```
ColdenCullen: This reads weirdly. How about "There are currently no open subscriptions, resubscribe is unnecessary and no-op."

PR #101, opened 2019-11-02
Updated tests to use mem tracer instead of test allocator

justinboswell:

### DIFF
--- a/tests/aws_iot_client_test.c
+++ b/tests/aws_iot_client_test.c
@@ -238,14 +238,14 @@ int main(int argc, char **argv) {
     const char *private_key = argv[3];
     const char *root_ca = argv[4];
 
-    AWS_TEST_ALLOCATOR_INIT(iot_client);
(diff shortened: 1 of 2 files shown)


tests/aws_iot_client_test.c
```diff
@@ -238,14 +238,14 @@ int main(int argc, char **argv) {
     const char *private_key = argv[3];
     const char *root_ca = argv[4];
 
-    AWS_TEST_ALLOCATOR_INIT(iot_client);
+    struct aws_allocator *allocator = aws_mem_tracer_new(aws_default_allocator(), AWS_MEMTRACE_BYTES, 0);
```
JonathanHenson: Why isn’t this just setup as the test allocator? Then none of this is necessary

justinboswell: It IS the test allocator. These aren't tests, they're fuzzes.

PR #107, opened 2019-12-01
Reconnect task leak

bretambrose:
* fix reconnect task leaks if the initial connection attempt fails
* initial attempt at guarding against connection state race conditions - unfortunately the mqtt logic is really resistant to a "nice" wrapper where there are balance lock-unlock pairs.  Instead there are a ton of unlocks before early outs and you have to manually verify that all control paths unlock.
* move shutdown to client.c

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -77,7 +77,87 @@ void aws_mqtt_client_clean_up(struct aws_mqtt_client *client) {
     AWS_ZERO_STRUCT(*client);
 }
 
-static void s_mqtt_client_shutdown(
+/* Reconnect helpers */
+
+static void s_attempt_reconnect(struct aws_task *task, void *userdata, enum aws_task_status status) {
+
+    (void)task;
+
+    struct aws_mqtt_reconnect_task *reconnect = userdata;
+    struct aws_mqtt_client_connection *connection = aws_atomic_load_ptr(&reconnect->connection_ptr);
+
+    if (status == AWS_TASK_STATUS_RUN_READY && connection) {
+        /* If the task is not cancelled and a connection has not succeeded, attempt reconnect */
+
+        aws_high_res_clock_get_ticks(&connection->reconnect_timeouts.next_attempt);
+        connection->reconnect_timeouts.next_attempt += aws_timestamp_convert(
+            connection->reconnect_timeouts.current, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+
+        AWS_LOGF_TRACE(
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Attempting reconnect, if it fails next attempt will be in %" PRIu64 " seconds",
+            (void *)connection,
+            connection->reconnect_timeouts.current);
+
+        /* Check before multipying to avoid potential overflow */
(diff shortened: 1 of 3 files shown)


source/client.c
```diff
@@ -77,6 +77,38 @@ void aws_mqtt_client_clean_up(struct aws_mqtt_client *client) {
     AWS_ZERO_STRUCT(*client);
 }
 
+/* Reconnect helpers */
+static void s_attempt_reconnect(struct aws_task *task, void *userdata, enum aws_task_status status);
+
+int s_create_reconnect_task(struct aws_mqtt_client_connection *connection) {
+    AWS_ASSERT(!connection->reconnect_task);
```
JonathanHenson: you can add a string here if you'd like:
AWS_ASSERT(!connection->reconnect_task && "my string I want the console output to show")

source/client.c
```diff
@@ -77,6 +77,38 @@ void aws_mqtt_client_clean_up(struct aws_mqtt_client *client) {
...
+    }
+    aws_atomic_init_ptr(&connection->reconnect_task->connection_ptr, connection);
+    connection->reconnect_task->allocator = connection->allocator;
+    aws_task_init(&connection->reconnect_task->task, s_attempt_reconnect, connection->reconnect_task, "mqtt_reconnect");
+
+    return AWS_OP_SUCCESS;
+}
+
+static void s_destroy_reconnect_task(struct aws_mqtt_client_connection *connection) {
+    if (connection->reconnect_task) {
```
JonathanHenson: Is this a safe use of atomics? shouldn't you be doing an atomic fetch for the null check?

source/client.c
```diff
@@ -77,6 +77,38 @@ void aws_mqtt_client_clean_up(struct aws_mqtt_client *client) {
...
+        aws_atomic_store_ptr(&connection->reconnect_task->connection_ptr, NULL);
+
+        /* Cancel if scheduled */
+        if (connection->reconnect_task->task.timestamp > 0) {
+            aws_event_loop_cancel_task(
+                aws_channel_get_event_loop(connection->slot->channel), &connection->reconnect_task->task);
```
JonathanHenson: Is there always a channel here? Also why do you need the channel? you have the event loop group, you could even always use the same event loop for reconnect task scheduling.

JonathanHenson:
also, did you check the ping task to make sure it's safe in this context?

source/client.c
bretambrose: I can't find a cancel API anywhere but on the event loop or the task scheduler.

source/client.c
bretambrose: This is just a cut-and-paste of what existed before into a function so I can call it in two spots.

source/client.c
bretambrose: True, the slot may have been unlinked by the time this task runs.  May need to move the destroy call into the channel shutdown callback instead.  Unsure what you mean by the second part; I don't see a way to cancel a scheduled task by any way other than the event loop it is scheduled on or the raw task task scheduler.

bretambrose:
> 
> 
> also, did you check the ping task to make sure it's safe in this context?

It's usage is unchanged, so safety shouldn't be any worse at least.

PR #109, opened 2019-12-02
Memory leaks

bretambrose:
* Fix for reconnect task leak when initial connection fails
* Fix for disconnect io message leak when initial connection fails
* Fix for crash when we get two acks back from a send followed by a resend.

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -353,14 +353,12 @@ static void s_attempt_reconnect(struct aws_task *task, void *userdata, enum aws_
 void aws_create_reconnect_task(struct aws_mqtt_client_connection *connection) {
     if (connection->reconnect_task == NULL) {
         connection->reconnect_task = aws_mem_calloc(connection->allocator, 1, sizeof(struct aws_mqtt_reconnect_task));
-        if (!connection->reconnect_task) {
-            AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "id=%p: Failed to allocate reconnect task", (void *)connection);
-        } else {
-            aws_atomic_init_ptr(&connection->reconnect_task->connection_ptr, connection);
-            connection->reconnect_task->allocator = connection->allocator;
-            aws_task_init(
-                &connection->reconnect_task->task, s_attempt_reconnect, connection->reconnect_task, "mqtt_reconnect");
-        }
+        AWS_FATAL_ASSERT(connection->reconnect_task != NULL);
+
+        aws_atomic_init_ptr(&connection->reconnect_task->connection_ptr, connection);
+        connection->reconnect_task->allocator = connection->allocator;
+        aws_task_init(
+            &connection->reconnect_task->task, s_attempt_reconnect, connection->reconnect_task, "mqtt_reconnect");
     }
 }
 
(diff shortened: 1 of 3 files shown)


source/client_channel_handler.c
```diff
@@ -114,6 +114,9 @@ static int s_packet_handler_connack(
 
         MQTT_CLIENT_CALL_CALLBACK_ARGS(connection, on_resumed, connack.connect_return_code, connack.session_present);
     } else {
+
+        aws_create_reconnect_task(connection);
```
JonathanHenson: what if this fails?

source/client.c
```diff
@@ -1022,17 +1036,6 @@ int aws_mqtt_client_connection_connect(
         aws_byte_buf_clean_up(&connection->client_id);
     }
 
-    /* Create the reconnect task for use later (probably) */
-    AWS_ASSERT(!connection->reconnect_task);
-    connection->reconnect_task = aws_mem_calloc(connection->allocator, 1, sizeof(struct aws_mqtt_reconnect_task));
-    if (!connection->reconnect_task) {
```
graebm: down at the goto _failed labels of this function, there are 2 attempts in a row to free `connection->reconnect_task`. They need to be deleted

source/client_channel_handler.c
```diff
@@ -505,15 +508,28 @@ static int s_shutdown(
...
                 }
-                aws_channel_slot_send_message(slot, message, AWS_CHANNEL_DIR_WRITE);
+
+                if (aws_channel_slot_send_message(slot, message, AWS_CHANNEL_DIR_WRITE)) {
+                    AWS_LOGF_DEBUG(
+                        AWS_LS_MQTT_CLIENT,
+                        "id=%p: failed to send courteous disconnect io message",
+                        (void *)connection);
+                    aws_mem_release(message->allocator, message);
+                    /* fall through to done: */
```
graebm: just goto done.
Somone in the future might add code between here and done

source/client_channel_handler.c
```diff
@@ -835,15 +862,19 @@ static void s_mqtt_disconnect_task(struct aws_channel_task *channel_task, void *
...
-        /* If the reconnect_task isn't scheduled, free it */
-        if (connection->reconnect_task && !connection->reconnect_task->task.timestamp) {
-            aws_mem_release(connection->reconnect_task->allocator, connection->reconnect_task);
+        /* Cancel if scheduled */
+        if (connection->reconnect_task->task.timestamp > 0 && connection->slot && connection->slot->channel) {
+            aws_event_loop_cancel_task(
+                aws_channel_get_event_loop(connection->slot->channel), &connection->reconnect_task->task);
         }
 
+        aws_mem_release(connection->reconnect_task->allocator, connection->reconnect_task);
```
graebm: this won't work, you can't reliably cancel "channel tasks", they travel through too many queues (channel queue, event-loop queue, scheduler queue)
The previous approach that NULLed things out and let the task clean itself up was the better approach

source/client_channel_handler.c
```diff
@@ -795,6 +811,17 @@ void mqtt_request_complete(struct aws_mqtt_client_connection *connection, int er
 
     aws_mutex_unlock(&connection->outstanding_requests.mutex);
 
+    if (elem == NULL) {
```
graebm: remove the `AWS_ASSERT(elem);` above, now that we handle this

PR #111, opened 2019-12-05
Memory pool related leaks

bretambrose:
* Fixes a number of leaks (some may be only theoretically possible) related to not releasing messages on failures to aws_channel_slot_send_message() or the corresponding packet encode function.
* Fixes several leaks related to requests being uncompleted or completed but not yet removed when the connection is shut down.


source/client_channel_handler.c
```diff
@@ -602,8 +626,42 @@ static void s_request_timeout_task(struct aws_channel_task *task, void *arg, enu
...
+         * If the task is cancelled, assume safe shutdown is in progress.
+         * There are two distinct situations where requests end up with a cancelled task:
+         *   (1) The request has been completed but disconnect was begun before the task had a chance to run
+         *       normally.  We need a way to ensure that aws_memory_pool_release is called for this request.
+         *
+         *       We can do that by
+         *          (1) setting request->cancelled to true.  Then the hash table's value removal function will
+         *              ensure aws_memory_pool_release is called when the table is cleared during final disconnect.
+         *          (2) If we're uncomfortable (semantically) with a request that is both "completed" and "cancelled"
+         *              then we could alternatively call aws_memory_pool_release here.
```
graebm: This comment will confuse future generations. Be more clear which of these things we're doing, or remove the alternative explanation of how we could do it.

source/client_channel_handler.c
```diff
@@ -602,8 +626,42 @@ static void s_request_timeout_task(struct aws_channel_task *task, void *arg, enu
...
+            request->cancelled = true;
+            if (!request->completed) {
+                AWS_LOGF_DEBUG(
+                    AWS_LS_MQTT_CLIENT,
+                    "static: task id %p, completing with interrupt request for packet id "
+                    "%" PRIu16 ".",
+                    (void *)task,
+                    request->message_id);
+                if (request->on_complete) {
+                    request->on_complete(
```
graebm: should we be setting `request->completed = true`?
since we're firing the on_complete callback?

bretambrose: True, task shouldn't get rerun but its safer to be sure.

source/client_channel_handler.c
bretambrose: Yeah was probably better left as a review comment, but wanted people to read it and say which they preferred (if any).

source/mqtt.c
```diff
@@ -151,6 +151,9 @@ bool aws_mqtt_is_valid_topic_filter(const struct aws_byte_cursor *topic_filter)
             AWS_DEFINE_ERROR_INFO_MQTT(
                 AWS_ERROR_MQTT_NO_TOPICS_FOR_RESUBSCRIBE,
                 "Resubscribe not necessary, client is not subscribed to any topics."),
+            AWS_DEFINE_ERROR_INFO_MQTT(
+                AWS_ERROR_MQTT_REQUEST_INTERRUPTED,
+                "Mqtt operation interrupted by connection shutdown"),
```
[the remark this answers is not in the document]
bretambrose: That seems reasonable.

PR #114, opened 2019-12-13
Release the websocket request upon websocket shutdown.

JonathanHenson:

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -745,6 +745,11 @@ static void s_on_websocket_shutdown(struct aws_websocket *websocket, int error_c
 
     s_mqtt_client_shutdown(connection->client->bootstrap, error_code, channel, connection);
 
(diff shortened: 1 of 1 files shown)


source/client.c
```diff
@@ -745,6 +745,11 @@ static void s_on_websocket_shutdown(struct aws_websocket *websocket, int error_c
 
     s_mqtt_client_shutdown(connection->client->bootstrap, error_code, channel, connection);
 
+    if (connection->websocket.handshake_request) {
+        aws_http_message_release(connection->websocket.handshake_request);
+        connection->websocket.handshake_request = NULL;
+    }
```
graebm: The free should occur in `s_on_websocket_setup` (when the connection is succeeds/fails)

With the free here, we don't free the request in the case that the connection fails

PR #119, opened 2020-02-11
Connection testing harness

JonathanHenson:


tests/connection_state_test.c
```diff
@@ -0,0 +1,947 @@
...
+    aws_mutex_unlock(&state_test_data->lock);
+    aws_condition_variable_notify_one(&state_test_data->cvar);
+}
+
+static bool s_is_connection_interupted(void *arg) {
+    struct mqtt_connection_state_test *state_test_data = arg;
+    return state_test_data->connection_interupted;
+}
+
+static void s_wait_for_interupt_to_complete(struct mqtt_connection_state_test *state_test_data) {
```
justinboswell: interrupt

tests/connection_state_test.c
```diff
@@ -0,0 +1,947 @@
...
+    struct aws_socket_endpoint endpoint;
+    struct aws_socket *listener;
+    struct aws_mqtt_client mqtt_client;
+    struct aws_mqtt_client_connection *mqtt_connection;
+    struct aws_socket_options socket_options;
+    bool session_present;
+    bool connection_completed;
+    bool client_disconnect_completed;
+    bool server_disconnect_completed;
+    bool connection_interupted;
```
justinboswell: interrupted

Also, interrupt/resume can happen multiple times, should these be counters? For that matter, should connection_completed, subscribe_completed, and so on also be counters?

tests/connection_state_test.c
```diff
@@ -0,0 +1,947 @@
...
+    bool connection_resumed;
+    bool subscribe_completed;
+    int interuption_error;
```
justinboswell: interruption_error

tests/connection_state_test.c
```diff
@@ -0,0 +1,947 @@
...
+static void s_wait_for_publish(struct mqtt_connection_state_test *state_test_data) {
+    aws_mutex_lock(&state_test_data->lock);
+    aws_condition_variable_wait_pred(
+        &state_test_data->cvar, &state_test_data->lock, s_is_publish_received, state_test_data);
+    state_test_data->publishes_received = 0;
+    state_test_data->expected_publishes = 0;
+    aws_mutex_unlock(&state_test_data->lock);
+}
+
+static void s_on_suback(
```
justinboswell: Should we also track unsubacks? Probably right?

tests/connection_state_test.c
JonathanHenson: they get flipped after they're used. They're only used for semaphores so counters wouldn't buy us anything.

tests/connection_state_test.c
JonathanHenson: we're not really testing the server.

JonathanHenson: so the act of the test not hanging tests that the client does its job.

PR #123, opened 2020-02-18
Lower default MQTT keep-alive.

graebm:
AWS IoT maxes out at (and defaults to) 1200 seconds, so let's use that.

Issue raised here: https://github.com/awslabs/aws-crt-python/issues/128

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -43,8 +43,8 @@
 /* 3 seconds */
 static const uint64_t s_default_request_timeout_ns = 3000000000;
 
-/* 1 hour */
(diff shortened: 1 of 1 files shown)


source/client.c
```diff
@@ -43,8 +43,8 @@
 /* 3 seconds */
 static const uint64_t s_default_request_timeout_ns = 3000000000;
 
-/* 1 hour */
-static const uint16_t s_default_keep_alive_sec = 3600;
+/* 20 minutes */
```
justinboswell: "maximum supported by the IoT Core service as of 2/18/2020"

Issue #135, opened 2020-06-14
Use after free in MQTT client


kankri:
Modifying the MQTT sample program [basic_pub_sub](https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/v1.7.0/samples/mqtt/basic_pub_sub/main.cpp) like this:

```diff
diff --git a/samples/mqtt/basic_pub_sub/main.cpp b/samples/mqtt/basic_pub_sub/main.cpp
index 3eead2d..395011d 100644
--- a/samples/mqtt/basic_pub_sub/main.cpp
+++ b/samples/mqtt/basic_pub_sub/main.cpp
@@ -334,6 +335,9 @@ int main(int argc, char *argv[])
         connection->Subscribe(topic.c_str(), AWS_MQTT_QOS_AT_LEAST_ONCE, onPublish, onSubAck);
         conditionVariable.wait(uniqueLock);

+        connection->Subscribe(topic.c_str(), AWS_MQTT_QOS_AT_LEAST_ONCE, onPublish, onSubAck);
+        conditionVariable.wait(uniqueLock);
+
         while (true)
         {
             String input;
```

and running it causes this output:

```
>_build\samples\mqtt\basic_pub_sub\Debug\basic-pub-sub.exe --endpoint ep --cert cert.pem --key key.pem --ca_file ca_cert.pem --topic some/topic
Connecting...
Connection completed with return code 0
Subscribe on topic some/topic on packetId 1 Succeeded
Subscribe on topic <DD><DD><DD><DD><DD><DD><DD><DD><DD><DD> on packetId 2 Succeeded
Enter the message you want to publish to topic some/topic and press enter. Enter 'exit' to exit this program.
Connection interrupted with error aws-c-io: AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE, TLS (SSL) negotiation failed
Connection resumed
exit
Disconnect completed
```

In the second subscription message the topic name contains hex bytes 0xdd and there is an AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE error which doesn't happen with the unmodified sample program.

Enabling logging, the following is shown for the 1st subscription:

```
[TRACE] [mqtt-client] - id=00A47128: Attempting send of subscribe 1 (first attempt)
[DEBUG] [mqtt-topic-tree] - tree=00A471A8: Inserting topic filter some/topic into topic tree
[TRACE] [mqtt-topic-tree] - action=04AAF310: Created action
[...]

PR #137, opened 2020-06-18
new string created in topic tree

TingDaoK:
*Issue #, if available:*
https://github.com/awslabs/aws-c-mqtt/issues/135

*Description of changes:*
- subscribe_task_topic takes ownership of the topic string. And the nodes in the topic tree will create a new string now, which means the caller of the aws_mqtt_topic_tree_insert() will need to clean up the topic_filter.

### DIFF
--- a/source/topic_tree.c
+++ b/source/topic_tree.c
@@ -39,7 +39,7 @@ struct topic_tree_action {
         AWS_MQTT_TOPIC_TREE_REMOVE,
     } mode;
 
-    /* All Modes */
+    /* All NODES */
     struct aws_mqtt_topic_node *node_to_update;
 
     /* ADD/UPDATE */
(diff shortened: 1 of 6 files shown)


source/topic_tree.c
```diff
@@ -368,8 +368,13 @@ static void s_topic_tree_action_commit(struct topic_tree_action *action, struct
                 action->node_to_update->topic = action->topic;
             }
             if (action->topic_filter) {
-                action->node_to_update->topic_filter = action->topic_filter;
-                action->node_to_update->owns_topic_filter = true;
+                if (action->node_to_update->owns_topic_filter && action->node_to_update->topic_filter) {
```
DavidOgunsAWS, line 371: I don't see how owns_topic_filter is safe to read then write here. Maybe I haven't fully understood the context, but it appears this is the point in code to authoritatively set the flag and not read its current value to make a decision.

TingDaoK: yeah, this is a weird situation. I did this since when there are two pending actions with the same topic filter in one commit. The idea was if the `node_to_update` already owned the `topic_filter`, the other insertion of the same `topic_filter` will be cleaned up in [aws_mqtt_topic_tree_transaction_insert](https://github.com/awslabs/aws-c-mqtt/pull/137/files#diff-2e434658c1666944abda6141f0b9400dL673). 
However, if the old action has not been committed yet. The old node will not own `topic_filter` until the commit called, the new string will not be cleaned up in `aws_mqtt_topic_tree_transaction_insert`. Possible memory leak here in that case.
It should be safe to read the `node_to_update` here, since the node will be zero out when it's created in ADD case, and the old node is always valid in UPDATE case.

PR #141, opened 2020-06-30
Multiple sub bug

TingDaoK:
*Issue #, if available:*

- topics is `/* list of subscribe_task_topic *s */`, but  we are pushing back the `aws_mqtt_topic_subscription`.

*Description of changes:*
- Expose the `subscribe_task_topic` struct, otherwise the `on_suback.multi` cannot deal with the topics they got...
- Or we need to rework the whole on_suback callback, let's take an easy fix for now.
- Apparently we still need some unit test about it, but that will be in TODO.

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -1187,6 +1187,15 @@ int aws_mqtt_client_connection_disconnect(
  * Subscribe
  ******************************************************************************/
 
+/* The lifetime of this struct is the same as the lifetime of the subscription */
+struct subscribe_task_topic {
+    struct aws_mqtt_client_connection *connection;
+
+    struct aws_mqtt_topic_subscription request;
+    struct aws_string *filter;
+    bool is_local;
+};
+
 /* The lifetime of this struct is from subscribe -> suback */
 struct subscribe_task_arg {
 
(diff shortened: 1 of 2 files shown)


TingDaoK:
Apparently, some regression test is needed. But since it's not a high priority thing... I'll see if I can finish other stuff and come back later.

TingDaoK:
> It looks like C++ is the only CRT that even tries to support multi-sub

Yeah, the fix for the C++ side is there https://github.com/awslabs/aws-crt-cpp/pull/146/files. And I'll add test for this fix.

TingDaoK:
Changed mind for the way of fixing it after fully reading the documentation. `struct aws_mqtt_topic_subscription` is for `/** Passed to subscribe() and suback callbacks */`. 
- Let users just deal with the same struct in callbacks as they passed into the subscribe() function call, which seems more user-friendly.
- Tests for this will be made shortly after this merged, in a sperate PR with other tests.

source/client.c
```diff
@@ -1429,7 +1442,7 @@ uint16_t aws_mqtt_client_connection_subscribe_multiple(
             AWS_BYTE_CURSOR_PRI(task_topic->request.topic));
 
         /* Push into the list */
-        aws_array_list_push_back(&task_arg->topics, &request);
+        aws_array_list_push_back(&task_arg->topics, &task_topic);
```
TingDaoK, line 1445: the array list topics should have pointers of `struct subscribe_task_topic`

source/client.c
```diff
@@ -1357,7 +1357,20 @@ static void s_subscribe_complete(
...
-        task_arg->on_suback.multi(connection, packet_id, &task_arg->topics, error_code, task_arg->on_suback_ud);
+        /* create a list of aws_mqtt_topic_subscription pointers from topics for the callback */
+        size_t list_len = aws_array_list_length(&task_arg->topics);
+        AWS_VARIABLE_LENGTH_ARRAY(uint8_t, cb_list_buf, list_len * sizeof(void *));
+        struct aws_array_list cb_list;
+        aws_array_list_init_static(&cb_list, cb_list_buf, list_len, sizeof(void *));
+        int err = 0;
+        for (size_t i = 0; i < list_len; i++) {
+            err |= aws_array_list_get_at(&task_arg->topics, &topic, i);
+            struct aws_mqtt_topic_subscription *subscription = &topic->request;
```
TingDaoK, line 1368: take the `struct aws_mqtt_topic_subscription` pointer to create an new list for the callback.

PR #146, opened 2020-07-14
Tests for public apis in client.h

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- Tests for `aws_mqtt_client_connection_set_will` `aws_mqtt_client_connection_set_login` `aws_mqtt_client_connection_set_on_any_publish_handler` `aws_mqtt_client_connection_subscribe_multiple` `aws_mqtt_client_connection_unsubscribe` The others APIs either never binded, or websockets.
- Easy fix for set_will & set_login function
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #136, opened 2020-06-17
Access violation in mqtt_create_request()


kankri:
Modifying the MQTT sample program [basic_pub_sub](https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/v1.7.1/samples/mqtt/basic_pub_sub/main.cpp) like this:

```diff
diff --git a/samples/mqtt/basic_pub_sub/main.cpp b/samples/mqtt/basic_pub_sub/main.cpp
index 3eead2d..bd9e6cc 100644
--- a/samples/mqtt/basic_pub_sub/main.cpp
+++ b/samples/mqtt/basic_pub_sub/main.cpp
@@ -78,6 +78,7 @@ int main(int argc, char *argv[])
      * Do the global initialization for the API.
      */
     ApiHandle apiHandle;
+    apiHandle.InitializeLogging(Aws::Crt::LogLevel::Trace, stderr);

     String endpoint;
     String certificatePath;
@@ -331,8 +332,11 @@ int main(int argc, char *argv[])
             conditionVariable.notify_one();
         };

-        connection->Subscribe(topic.c_str(), AWS_MQTT_QOS_AT_LEAST_ONCE, onPublish, onSubAck);
-        conditionVariable.wait(uniqueLock);
+        for (int i=0; i < 100;++i) {
+            printf("subscribe %d\n", i);
+            connection->Subscribe(topic.c_str(), AWS_MQTT_QOS_AT_LEAST_ONCE, onPublish, onSubAck);
+            conditionVariable.wait(uniqueLock);
+        }

         while (true)
         {
```

and running it:

```
_build\samples\mqtt\basic_pub_sub\Debug\basic-pub-sub.exe --endpoint ep --cert cert.pem --key key.pem --ca_file ca_cert.pem --topic some/topic --client_id clid
```

causes an access violation in `mqtt_create_request()` at [line 866](https://github.com/awslabs/aws-c-mqtt/blob/v0.4.31/source/client_channel_handler.c#L866):

```C
    } else if (aws_channel_thread_is_callers_thread(connection->slot->channel)) {
```

or [line 879](https://github.com/awslabs/aws-c-mqtt/blob/v0.4.31/source/client_channel_handler.c#L879):

```C
        aws_channel_schedule_task_now(connection->slot->channel, &next_request->timeout_task);
```

because `connection->slot` is nullptr.

[...]

PR #147, opened 2020-07-22
Connection revamp

TingDaoK:
## Issue
https://github.com/awslabs/aws-c-mqtt/issues/136
The state of connection is not thread-safe

## Solution
Taking a lock to protect the state from multiple threads.
But, we already have three locks for connection. Instead of introducing the fourth lock, revamp the connection structure to share the same lock.

## Changes made
- thread_data & synced_data as we did for aws-c-http. Using state of connection to product the configurations of connection.
- fix error for resubscribe. Similar to [Multiple sub bug](https://github.com/awslabs/aws-c-mqtt/pull/141). It's probably better to be in [tests](https://github.com/awslabs/aws-c-mqtt/pull/146). Involved some thread-safe fixes as well, so keep it here.
- Left a tons of todos, most of them are about the retry police & QoS 2. We will finish them in the near future?
- Documentation rewrite. Github readme needs update as well.
- Make sure the on_completed for requests function will be called in all situation.

## TODOs
- Retry police.
- Make sure Websockets is thread-safe.
- We need to do the right thing for QoS 2 publish. If we want to fully support QoS 2.
- update github readme


tests/connection_state_test.c
```diff
@@ -915,3 +915,20 @@ AWS_TEST_CASE_FIXTURE(
     s_test_mqtt_connection_offline_publish_fn,
     s_clean_up_mqtt_server_fn,
     &test_data)
+
+/**
+ * TODO: Another race condition may cause the publish results differently due to our inconsistent retry police.
```
TingDaoK, line 920: A workaround is possible, but if we are going to make change in the near future (next week), I will wait until the change to make this test.

include/aws/mqtt/private/client_impl.h
```diff
@@ -85,77 +88,100 @@ struct aws_mqtt_client_connection {
...
+    struct aws_channel_handler handler;
+    struct aws_channel_slot *slot;
+
+    /* The host information, changed by user when state is AWS_MQTT_CLIENT_STATE_DISCONNECTED */
     struct aws_string *host_name;
     uint16_t port;
     struct aws_tls_connection_options tls_options;
     struct aws_socket_options socket_options;
 
+    /* Connect parameters */
```
TingDaoK, line 101: All of those connect parameters are changed by user when the state is CONNECTED/DISCONNECTED, it will only be used by eventloop thread for next attempt of connecting, when the state is CONNECTING/RECONNECTING. So, we don't need a lock for them.

include/aws/mqtt/private/client_impl.h
```diff
@@ -85,77 +88,102 @@ struct aws_mqtt_client_connection {
...
+        struct aws_hash_table outstanding_requests_table;
+
+        /**
+         * List of all requests that cannot be scheduled until the connection comes online.
+         * TODO: make the size of the list configurable. Stop it from from ever-growing
+         */
+        struct aws_linked_list pending_requests_list;
+    } synced_data;
+
+    /* TODO: websocket is thread-safe? If not, make it thread-safe. */
```
justinboswell, line 186: The websocket API should be thread safe, same as HTTP.

include/aws/mqtt/private/client_impl.h
```diff
@@ -85,77 +88,102 @@ struct aws_mqtt_client_connection {
...
+        struct aws_byte_buf payload;
+    } will;
+    struct {
+        uint64_t current;      /* seconds */
+        uint64_t min;          /* seconds */
+        uint64_t max;          /* seconds */
+        uint64_t next_attempt; /* milliseconds */
+    } reconnect_timeouts;
     /* User connection callbacks */
     aws_mqtt_client_on_connection_complete_fn *on_connection_complete;
```
justinboswell, line 121: These get set by the connect call, what if you connect while a connection/disconnection is in flight? Wouldn't these possibly produce a race? Should they be in the synced data?

include/aws/mqtt/private/client_impl.h
```diff
@@ -85,77 +88,102 @@ struct aws_mqtt_client_connection {
...
-    enum aws_mqtt_client_connection_state state;
-
-    /* Channel handler information */
-    struct aws_channel_handler handler;
-    struct aws_channel_slot *slot;
-
-    /* Keeps track of all open subscriptions */
-    struct aws_mqtt_topic_tree subscriptions;
     aws_mqtt_client_publish_received_fn *on_any_publish;
     void *on_any_publish_ud;
```
justinboswell, line 128: This also feels like it should be synced, because delivery can be happening in one thread while this callback is set from another.

include/aws/mqtt/private/client_impl.h
```diff
@@ -85,77 +88,102 @@ struct aws_mqtt_client_connection {
...
-        struct aws_mutex mutex;
-    } outstanding_requests;
-    /* List of all requests that cannot be scheduled until the connection comes online */
-    struct {
-        struct aws_linked_list list;
-        struct aws_mutex mutex;
-    } pending_requests;
+    /* Connection tasks. */
     struct aws_mqtt_reconnect_task *reconnect_task;
     struct aws_channel_task ping_task;
```
justinboswell, line 134: These are both event loop local, right?

include/aws/mqtt/private/client_impl.h
```diff
@@ -85,77 +88,102 @@ struct aws_mqtt_client_connection {
...
 
+    /**
+     * Number of times this connection has successfully CONNACK-ed, used
+     * to ensure on_connection_completed is sent on the first completed
+     * CONNECT/CONNACK cycle
+     */
+    size_t connection_count;
```
justinboswell, line 141: should this be synchronized?

source/client.c
```diff
@@ -80,96 +101,106 @@ static void s_mqtt_client_shutdown(
 
     AWS_LOGF_TRACE(
         AWS_LS_MQTT_CLIENT, "id=%p: Channel has been shutdown with error code %d", (void *)connection, error_code);
+    enum aws_mqtt_client_connection_state ori_state;
```
justinboswell, line 104: Maybe ```prior_state```? ```prev_state```?

source/client.c
```diff
@@ -80,96 +101,106 @@ static void s_mqtt_client_shutdown(
...
+        case AWS_MQTT_CLIENT_STATE_CONNECTING:
             AWS_LOGF_TRACE(
+                AWS_LS_MQTT_CLIENT, "id=%p: Initial connection attempt failed, calling callback", (void *)connection);
+            MQTT_CLIENT_CALL_CALLBACK_ARGS(connection, on_connection_complete, error_code, 0, false);
+            break;
+        case AWS_MQTT_CLIENT_STATE_CONNECTED: {
+            AWS_LOGF_DEBUG(
                 AWS_LS_MQTT_CLIENT,
-                "id=%p: Caller requested disconnect from on_interrupted callback, aborting reconnect",
+                "id=%p: Connection lost, calling callback and attempting reconnect",
```
justinboswell, line 175: Connection interrupted (we should try to be consistent so users understand that these are related)

source/client.c
```diff
@@ -430,38 +469,49 @@ struct aws_mqtt_client_connection *aws_mqtt_client_connection_new(struct aws_mqt
     /* Initialize the client */
     connection->allocator = client->allocator;
     connection->client = client;
-    connection->state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
+    connection->synced_data.state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
```
justinboswell, line 472: Should probably zero out the entire synced_data structure before we start filling in values

source/client.c
```diff
@@ -561,6 +621,25 @@ void aws_mqtt_client_connection_destroy(struct aws_mqtt_client_connection *conne
...
+static int s_check_connection_state_for_configuration(struct aws_mqtt_client_connection *connection) {
+    int result = AWS_OP_SUCCESS;
+    { /* BEGIN CRITICAL SECTION */
+        s_mqtt_connection_lock_synced_data(connection);
+
+        if (connection->synced_data.state != AWS_MQTT_CLIENT_STATE_DISCONNECTED &&
+            connection->synced_data.state != AWS_MQTT_CLIENT_STATE_CONNECTED) {
+            AWS_LOGF_WARN(
+                AWS_LS_MQTT_CLIENT,
+                "id=%p: The connection is not stable now. Not safe to configure the connection. Try again later.",
```
justinboswell, line 634: "Connection is currently pending connect/disconnect. Unable to make configuration changes until pending operation completes"

source/client.c
```diff
@@ -1800,37 +1929,53 @@ uint16_t aws_mqtt_client_connection_subscribe_local(
...
     struct subscribe_task_arg *task_arg = userdata;
     bool initing_packet = task_arg->subscribe.fixed_header.packet_type == 0;
     struct aws_io_message *message = NULL;
 
+    size_t sub_count = aws_mqtt_topic_tree_get_sub_count(&task_arg->connection->thread_data.subscriptions);
+    if (sub_count == 0) {
+        aws_raise_error(AWS_ERROR_MQTT_NO_TOPICS_FOR_RESUBSCRIBE);
+        AWS_LOGF_WARN(
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Not subscribed to any topics. Resubscribe is unnecessary, no packet will be sent. Error %s.",
```
justinboswell, line 1959: How would a user have caused this? Is there something else we should tell them, or should we just silently ignore this situation?

source/client_channel_handler.c
```diff
@@ -24,6 +24,18 @@
  * Packet State Machine
  ******************************************************************************/
 
+static void s_mqtt_connection_lock_synced_data(struct aws_mqtt_client_connection *connection) {
```
justinboswell, line 27: Maybe instead of having 2 copies of these functions, just put them in one place without static (probably client.c) and extern them to the other.

source/client_channel_handler.c
```diff
@@ -728,53 +762,57 @@ static void s_request_timeout_task(struct aws_channel_task *task, void *arg, enu
...
-            aws_linked_list_push_back(&request->connection->pending_requests.list, &request->list_node);
-            aws_mutex_unlock(&request->connection->pending_requests.mutex);
+                (void *)(connection),
+                request->packet_id);
+            { /* BEGIN CRITICAL SECTION */
+                s_mqtt_connection_lock_synced_data(connection);
+                aws_hash_table_remove(
+                    &connection->synced_data.outstanding_requests_table, &request->packet_id, &elem, &was_present);
+                /* If the request is not in outstanding_requests_table, that's a bug */
+                AWS_ASSERT(was_present);
```
justinboswell, line 779: This should be a warning. It shouldn't exit the program, since the state is already what we wanted: the table doesn't contain the request

tests/CMakeLists.txt
```diff
@@ -37,8 +37,11 @@ add_test_case(mqtt_connection_connack_timeout)
 add_test_case(mqtt_connect_subscribe)
 add_test_case(mqtt_connect_subscribe_multi)
 add_test_case(mqtt_connect_unsubscribe)
+# we probably need to deprecate resubscribe or not
+# add_test_case(mqtt_connect_resubscribe)
```
justinboswell, line 41: Just delete the test

include/aws/mqtt/private/client_impl.h
TingDaoK: aws_mqtt_client_connection_connect only be called when state is AWS_MQTT_CLIENT_STATE_DISCONNECTED. So if the state is AWS_MQTT_CLIENT_STATE_DISCONNECTED, there will not be connection/disconnection in flight. BTW the other connection configuration set from seperate functions will only be called as CONNECTED or DISCONNECTED, so, they should be safe.

include/aws/mqtt/private/client_impl.h
TingDaoK: yeah... But, we are not putting task into thread_data at HTTP... My understanding is it's probably straightforward.

include/aws/mqtt/private/client_impl.h
TingDaoK: It's set to zero by the `aws_mqtt_client_connection_connect`, when connection is still DISCONNECTED, and changed/used only by the `s_packet_handler_connack`, which is from the eventloop thread as CONACK received. So I think it's fine...

include/aws/mqtt/private/client_impl.h
TingDaoK: hmmm, I thought about this. I cannot remember how I convinced myself we don't need a lock for this... Anyway, now I agree that we need put it into the synced_data, will do

include/aws/mqtt/client.h
```diff
@@ -169,7 +172,14 @@ struct aws_mqtt_topic_subscription {
  *                           within this amount of time (milliseconds). If you specify 0, a default value of 3 seconds
  *                           is used. Alternatively, tcp keep-alive may be away to accomplish this in a more efficient
  *                           (low-power) scenario, but keep-alive options may not work the same way on every platform
- *                           and OS version. This duration must be shorter than keep_alive_time_secs.
+ *                           and OS version. This duration must be shorter than keep_alive_time_secs. It's also for
+ *                           reattempt other requests.
```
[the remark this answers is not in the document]
TingDaoK: I don't know... Do we decide to do that? For now, it will re-attempt requests includes subscribe/unsubscribe/publish with QoS > 0.

source/client.c
TingDaoK: The previous connection had no subscription, then the resubscribe called. Yeah, it's probably better to silently ignore this

include/aws/mqtt/client.h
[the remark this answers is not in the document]
justinboswell: Yeah, I was just rephrasing the comment. The intent is the same. Requests (at least in my head) includes sub/unsub/pub.

include/aws/mqtt/private/client_impl.h
TingDaoK: right! Here is the problem, user may probably call functions that grab a lock from callback, which will cause a deadlock... A workaround I can think of is do not allow user to change this on_publish callback when the state is CONNECTED.🤔

tests/CMakeLists.txt
TingDaoK: If we are going to keep the resubscribe function (which I think we can), I'll build the test recently...

include/aws/mqtt/private/client_impl.h
justinboswell: That's a perfectly good solution.

tests/CMakeLists.txt
TingDaoK: test built

PR #148, opened 2020-07-24
fix possible memory-leak when disconnect while reconnecting

TingDaoK:
*Issue #, if available:*

Pending request list never get cleaned up if connection never connects.
 
*Description of changes:*

- Make sure no requests will be made after disconnect called before the channel finishes shutting down. (which is discussed further at the second Debatable behavior)
- Stop creating the slot and sending the CONNECT if disconnect called. In case we never receive a CONACK, the disconnect can still succeed.
- Cleanup the pending_requests_list and reconnect_task

*Debatable behavior*

- We are canceling all the uncompleted requests (subscribe/unsubscribe/publish with all QoS) if user calls disconnect. And we will keep them when connection interrupted. Regardless of the clean_session... Seems like should be part of the retry policy we need to discuss.

- *Interesting one*. If the request is made after the disconnect called, before the channel finishes shutting down, it will be canceled. However, if the request is made after the channel finishes shutting down, it will be retried once the same connection connects again. Ah, I love it
Probably just werid use case... User should not do this kind of stuff? Like `It is not safe to make request after on_disconnect callback fired`. We can enforce this behavior by verifying the state is not DISCONNECTING. (This's what we did in the PR)


source/client.c
```diff
@@ -257,6 +258,21 @@ static void s_mqtt_client_init(
...
 
+    /* user request disconnect before the channel setted up, stop installing the slot and sending CONNECT. */
+    bool is_closing = false;
+
+    { /* BEGIN CRITICAL SECTION */
+        s_mqtt_connection_lock_synced_data(connection);
+        is_closing = connection->synced_data.state == AWS_MQTT_CLIENT_STATE_DISCONNECTING;
+        s_mqtt_connection_unlock_synced_data(connection);
+    } /* END CRITICAL SECTION */
+    if (is_closing) {
```
bretambrose, line 269: to be honest, I don't see what good this does.  the client could easily be set to disconnecting right after this check is made.

include/aws/mqtt/mqtt.h
```diff
@@ -51,6 +51,8 @@ enum aws_mqtt_error {
     AWS_ERROR_MQTT_UNEXPECTED_HANGUP,
     AWS_ERROR_MQTT_NO_TOPICS_FOR_RESUBSCRIBE,
     AWS_ERROR_MQTT_CONNECTION_SHUTDOWN,
+    AWS_ERROR_MQTT_CONNECTION_DESTORIED,
```
justinboswell, line 54: DESTROYED?

source/mqtt.c
```diff
@@ -143,7 +143,13 @@ bool aws_mqtt_is_valid_topic_filter(const struct aws_byte_cursor *topic_filter)
                 "Resubscribe not necessary, client is not subscribed to any topics."),
             AWS_DEFINE_ERROR_INFO_MQTT(
                 AWS_ERROR_MQTT_CONNECTION_SHUTDOWN,
-            "Mqtt operation interrupted by connection shutdown"),
+                "Mqtt operation interrupted by connection shutdown."),
```
justinboswell, line 146: MQTT

source/mqtt.c
```diff
@@ -143,7 +143,13 @@ bool aws_mqtt_is_valid_topic_filter(const struct aws_byte_cursor *topic_filter)
...
+            AWS_DEFINE_ERROR_INFO_MQTT(
+                AWS_ERROR_MQTT_CONNECTION_DESTORIED,
+                "Destory connection has been called, all uncompleted requests will fail."),
```
justinboswell, line 149: Destroy/DESTROYED

TingDaoK:
> Once disconnect has been called by the user, we should not accept any further requests, it looks to me like you've handled that by reporting that the connection as DISCONNECTING, which is just fine. For connection destroyed though, how does that really differ from SHUTDOWN?

Oh, sorry that I forgot to update the write-up... I have changed mind during writing the "behavior", it should be "Make sure no requests will be made after disconnect called before the channel finishes shutting down". I updated now. 

So the channel shutdown async,  it will happen after the `aws_mqtt_client_connection_disconnect` called by user. And on_disconnect will be fired as the channel finishes shutting down (state of connection will be switched from AWS_MQTT_CLIENT_STATE_DISCONNECTING to AWS_MQTT_CLIENT_STATE_DISCONNECTED). aws_mqtt_client_connection_destroy can only be called on a disconnected connection, and it will clean up all the resource. To summarize, the destroy just clean up the resource and the SHUTDOWN is the result of channel after aws_mqtt_client_connection_disconnect.

TingDaoK:
> Does this PR rely on the connection revamp?

Yes, this PR is merging into https://github.com/awslabs/aws-c-mqtt/pull/147

source/mqtt.c
TingDaoK: I want to mean `aws_mqtt_client_connection_destroy` is called...

source/client.c
```diff
@@ -102,6 +102,8 @@ static void s_mqtt_client_shutdown(
     AWS_LOGF_TRACE(
         AWS_LS_MQTT_CLIENT, "id=%p: Channel has been shutdown with error code %d", (void *)connection, error_code);
     enum aws_mqtt_client_connection_state ori_state;
+    struct aws_linked_list pending_requests_to_clear;
+    aws_linked_list_init(&pending_requests_to_clear);
```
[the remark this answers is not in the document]
TingDaoK: Oh, nice catch... I changed the implementation to clear the pending requests while destroying the connection now, but forgot to remove this list...

source/client.c
TingDaoK: The point is the `mqtt_disconnect_impl` will do nothing if the slot is not installed. As long as the state is set to disconnecting (`mqtt_disconnect_impl` happens) after the slot is installed, we will be fine.

bretambrose: or just extend the lock until then maybe?

justinboswell: Yeah, this is the only confusing bit left, what is the rationale here?

justinboswell:
g2g except for that one section Bret called out

source/client.c
TingDaoK: The reason for this early stop here is that if the slot is not installed and user request to disconnect, the connection will not be closed from our side. We will wait for the broker/server side to close the connection, which is different from what user wants. 
So, we need to ensure the disconnect happens before the slot installed will fall into the early stop and connection is closed from our side as user requested.

PR #149, opened 2020-08-11
Ref

bretambrose:
* rework ref counting and asynchronous shutdown for a number of core types and their downstream dependents

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -56,36 +56,47 @@ void mqtt_connection_unlock_synced_data(struct aws_mqtt_client_connection *conne
     (void)err;
 }
 
-/*******************************************************************************
- * Client Init
- ******************************************************************************/
-
-int aws_mqtt_client_init(
-    struct aws_mqtt_client *client,
-    struct aws_allocator *allocator,
-    struct aws_client_bootstrap *bootstrap) {
-
-    aws_mqtt_fatal_assert_library_initialized();
-
-    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "client=%p: Initalizing MQTT client", (void *)client);
-
-    AWS_ZERO_STRUCT(*client);
-    client->allocator = allocator;
-    client->bootstrap = aws_client_bootstrap_acquire(bootstrap);
-
-    return AWS_OP_SUCCESS;
-}
-
-/*******************************************************************************
- * Client Clean Up
- ******************************************************************************/
-
-void aws_mqtt_client_clean_up(struct aws_mqtt_client *client) {
+static void s_aws_mqtt_client_destroy(struct aws_mqtt_client *client) {
 
     AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "client=%p: Cleaning up MQTT client", (void *)client);
     aws_client_bootstrap_release(client->bootstrap);
 
-    AWS_ZERO_STRUCT(*client);
+    aws_mem_release(client->allocator, client);
+}
+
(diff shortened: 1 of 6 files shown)


source/client.c
```diff
@@ -501,8 +513,112 @@ static void s_outstanding_request_destroy(void *item) {
...
+    bool call_destroy_final = false;
+
+    { /* BEGIN CRITICAL SECTION */
+        mqtt_connection_lock_synced_data(connection);
+        if (connection->synced_data.state != AWS_MQTT_CLIENT_STATE_DISCONNECTED) {
+
+            /*
+             * We don't call the on_disconnect callback until we've transitioned to the DISCONNECTED state.  So it's
+             * safe to change it now while we hold the lock since we know we're not DISCONNECTED yet.
+             */
```
TingDaoK, line 602: I know I said it should be safe to do this before... 
However, I just realized a crazy corner case. The problem is we are not protecting the callback by a lock, since we don't want to invoke any callback with holding a lock.  So, it's very unlikely, but possible that if the state is DISCONNECTED, `aws_mqtt_client_connection_connect` called and switch the state to CONNECTING before the `on_disconnect` invoked. We will crash...
We can probably document it like "don't call connect if you call disconnect and the callback has not been invoked"?

source/client.c
```diff
@@ -598,90 +716,18 @@ struct aws_mqtt_client_connection *aws_mqtt_client_connection_new(struct aws_mqt
...
-        aws_tls_connection_options_clean_up(&connection->websocket.proxy->tls_options);
+    return connection;
+}
 
-        aws_mem_release(connection->allocator, connection->websocket.proxy);
-        connection->websocket.proxy = NULL;
-        connection->websocket.proxy_options = NULL;
+void aws_mqtt_client_connection_release(struct aws_mqtt_client_connection *connection) {
+    if (connection != NULL) {
+        aws_ref_count_release(&connection->ref_count);
```
TingDaoK, line 729: Just want to make sure, because the destroy of connection is asynced call. So it is possible to call `aws_ref_count_release` when the refcount is zero already but the connection is still not NULL, right? It will lead to a crash. Is that what we want?

TingDaoK: If we want it to crash for negative refcount, can we do `ASSERT(connection)` to make it more consistent?

bretambrose: Well sure it's possible but that's 100% user error and not something I think we need to worry about.

source/client.c
bretambrose: I don't think this is a case we need to worry about.  We're only executing this logic if every single reference to the connection has been released.  Nothing stop the user from releasing and then using the pointer but that's a usage error on them, not us.

source/client.c
TingDaoK: Would it be better to `ASSERT(connection)`, in which case, the invalid call off `aws_mqtt_client_connection_release ` will always lead to a crash. For now, it may crash or silently do nothing.

TingDaoK (approved):

source/client.c
bretambrose: All of the release functions check null, which makes it easier to write unconditional tear-down functions that work regardless of partial construction (i.e. got halfway through _new() and failed).

bretambrose: aws_mqtt_connection_new has the old approach which has many goto-destinations, one for each thing that needs to be cleaned up and everything has to be perfectly aligned.  More recently, we just make more resilient _destroy or _release functions and call it regardless of how far we got in construction

PR #152, opened 2020-09-18
Configurable auto reconnect

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- Add stop_auto_reconnect to configuration of connection. 
- Making the first attempt of connecting no different from the following attempts
- ping_timeout_ms -> request_timeout_ms, which will affect the timeout for PING/CONNECT. Note: for now, it still affects the other requests, which will be removed with the following up PRs.
- Trivial updates for documentations

Alternative:
For stop_auto_reconnect, if a connected connection lost, we are invoking the on_interrupted callback for now. It may be to have a required sperate on_connection_lost callback in this case? I don't have strong preference.
- Can avoid a case that user don't know what's happening when the connection lost. Since on_interrupted is not required
- But, also think it may be the requirement for user/CRTs. So, stay with the old on_interrupted callback for now

TODO:
CONNACK result code should lead to a different behavior.


source/client.c
```diff
@@ -175,38 +193,45 @@ static void s_mqtt_client_shutdown(
...
+                    AWS_LS_MQTT_CLIENT,
+                    "id=%p: Connection interrupted, calling callback and attempting reconnect",
+                    (void *)connection);
+                MQTT_CLIENT_CALL_CALLBACK_ARGS(connection, on_interrupted, error_code);
+
+                /* In case user called disconnect from the on_interrupted callback */
+                bool stop_reconnect;
+                { /* BEGIN CRITICAL SECTION */
+                    mqtt_connection_lock_synced_data(connection);
+                    stop_reconnect = connection->synced_data.state == AWS_MQTT_CLIENT_STATE_DISCONNECTING;
```
justinboswell, line 209: Should we also set `connection->auto_reconnect` to `false` somewhere in here?

justinboswell (approved):

source/client.c
TingDaoK: In this case, there will be no further reconnect happening for this connection object. To reconnect/connect, user will need to call `aws_mqtt_client_connection_connect` with new option. So, it's not necessary to change the `connection->auto_reconnect` here.

TingDaoK:
Reworked at https://github.com/awslabs/aws-c-mqtt/pull/159

PR #153, opened 2020-09-22
check the return code before on_completed callback

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/client_channel_handler.c
+++ b/source/client_channel_handler.c
@@ -93,8 +93,9 @@ static int s_packet_handler_connack(
         }
 
         was_reconnecting = connection->synced_data.state == AWS_MQTT_CLIENT_STATE_RECONNECTING;
-        connection->synced_data.state = AWS_MQTT_CLIENT_STATE_CONNECTED;
         if (connack.connect_return_code == AWS_MQTT_CONNECT_ACCEPTED) {
+            /* Don't change the state if it's not ACCEPTED by broker */
+            connection->synced_data.state = AWS_MQTT_CLIENT_STATE_CONNECTED;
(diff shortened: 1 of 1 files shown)


source/client_channel_handler.c
```diff
@@ -160,10 +136,37 @@ static int s_packet_handler_connack(
             "id=%p: invalid connect return code %d, disconnecting",
             (void *)connection,
             connack.connect_return_code);
-        /* If error code returned, disconnect */
+        /* If error code returned, disconnect, on_completed will be invoked from shutdown process */
```
bretambrose, line 139: Can you link where this is done?

TingDaoK: https://github.com/awslabs/aws-c-mqtt/blob/master/source/client.c#L181
The return code will not be shown to the client. Only the error code `AWS_ERROR_MQTT_PROTOCOL_ERROR` will be there

bretambrose: Nice.

bretambrose (approved):

source/client_channel_handler.c
```diff
@@ -160,10 +136,37 @@ static int s_packet_handler_connack(
...
         aws_channel_shutdown(connection->slot->channel, AWS_ERROR_MQTT_PROTOCOL_ERROR);
```
graebm, line 140: This isn't a protocol error. We should really declare distinct error codes for each of the CONNACK non-zeroes.
AWS_ERROR_MQTT_CONNECTION_REFUSED_UNACCEPTABLE_PROTOCOL_VERSION???

PR #154, opened 2020-09-30
Retry of requests

TingDaoK:
Rework the retry behavior for requests:
1. Not retry publish with QoS 0
2. Introduce the limit for the offline queue (removed)
3. Not queuing publish QoS 0 and PINGREQ when offline
4. Whether the pakcet is sent or not, the retriable packet will be queued when connection lost, and the sent packets will be resent in the same order when they are sent before. Following requirement from [Spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105)
5. Will not resent the packets on a health connection, even if the response is not received for a long time.
6. [Spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718103) only **REQUIRED** to retry publish with QoS > 0, we will retry subscribe/unsubscribe as well to keep the consistent of our behavior. 

TODO:
1. Behavior about reconnect. Respect the clean_session flag. -> it'll be a separate PR. This one is already very complicated. 
2. To ensure, what kind of "internal" error needs to be retryed. -> Need some discussion.


source/client_channel_handler.c
```diff
@@ -820,34 +775,58 @@ uint16_t mqtt_create_request(
...
             AWS_LOGF_ERROR(
-                AWS_LS_MQTT_CLIENT, "id=%p: Disconnect requested, stop creating any new request.", (void *)connection);
+                AWS_LS_MQTT_CLIENT,
+                "id=%p: Disconnect requested, stop creating any new request until disconnect process finishes.",
+                (void *)connection);
             aws_raise_error(AWS_ERROR_MQTT_CONNECTION_DISCONNECTING);
             return 0;
         }
 
+        if (noRetry && connection->synced_data.state != AWS_MQTT_CLIENT_STATE_CONNECTED) {
```
TingDaoK, line 799: Options:
1. invoke complete callback & clean it up during the task. Problem is the channel may not exist, it will still be a synced call.
2. Fail the call. Exception with no offline queueing for QoS 0.
3. For ping, it's not an error and no complete callback, it's fine to do this. But, for publish, this will not invoke the on_complete callback, it should be consider as error?

include/aws/mqtt/private/client_impl.h
```diff
@@ -117,6 +124,12 @@ struct aws_mqtt_client_connection {
         uint64_t max;          /* seconds */
         uint64_t next_attempt; /* milliseconds */
     } reconnect_timeouts;
+    /**
+     * Length of the pending list. When the number of pending request reach this number, the oldest request will be
+     * ejected from the list, and completed with failure. Set it to zero to disable offline queueing.
+     */
+    size_t pending_list_len;
```
TingDaoK, line 131: Should we add a separate `set` API for this? Hmmm, seems useful to me.

source/client.c
```diff
@@ -1651,6 +1651,8 @@ static enum aws_mqtt_client_request_state s_subscribe_send(uint16_t packet_id, b
 
     /* This is not necessarily a fatal error; if the subscribe fails, it'll just retry. Still need to clean up though.
      */
+    /* TODO: NOT really agree on retry for this failure, it's reasonable to retry if the error is socket_closed or not
+     * connected, otherwiseit's probably a program error. */
```
TingDaoK, line 1655: At least, if the socket is closed/not connected, it's fine to put it as on_going, which will be moved to the pending list when the channel finishes shutdown process. Otherwise, it's should be more complicated to deal with.

source/client_channel_handler.c
```diff
@@ -626,6 +626,7 @@ void aws_mqtt_internal_complete_request(struct aws_mqtt_request *request, int er
     /* Fire the callback and clean up the memory, as the connection get destoried. */
     if (!request->completed) {
         if (request->on_complete) {
+            /* TODO: I am invoking the callback with lock hold XXXXXX */
             request->on_complete(connection, request->packet_id, error_code, request->on_complete_ud);
```
TingDaoK: Is it safe to do the following?
```
mqtt_connection_unlock_synced_data(connection);
request->on_complete(connection, request->packet_id, error_code, request->on_complete_ud);
mqtt_connection_lock_synced_data(connection);
```

TingDaoK:
This CI failure is super weird. `s_mqtt_client_shutdown`, which should happen from the eventloop thread, and `aws_mqtt_client_connection_publish`, which should happen from the main thread, are having a deadlock. I really don't know... It makes no sense to me...

source/client_channel_handler.c
TingDaoK: NOPE, CANNOT DO THIS! [It](https://github.com/awslabs/aws-c-mqtt/pull/157/checks?check_run_id=1233344391) almost killed me!!!

source/client_channel_handler.c
graebm: don't invoke completion callback synchonously

source/client.c
```diff
@@ -47,12 +47,16 @@ static int s_mqtt_client_connect(
 void mqtt_connection_lock_synced_data(struct aws_mqtt_client_connection *connection) {
     int err = aws_mutex_lock(&connection->synced_data.lock);
     AWS_ASSERT(!err);
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: Lock hold", (void *)connection);
```
graebm, line 50: trivial: remove

source/client.c
```diff
@@ -47,12 +47,16 @@ static int s_mqtt_client_connect(
...
     (void)err;
 }
 
 void mqtt_connection_unlock_synced_data(struct aws_mqtt_client_connection *connection) {
+    ASSERT_SYNCED_DATA_LOCK_HELD(connection);
+
     int err = aws_mutex_unlock(&connection->synced_data.lock);
     AWS_ASSERT(!err);
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: Lock released", (void *)connection);
```
graebm, line 59: trivial: remove

source/client.c
```diff
@@ -116,6 +120,19 @@ static void s_mqtt_client_shutdown(
     enum aws_mqtt_client_connection_state prev_state;
     { /* BEGIN CRITICAL SECTION */
         mqtt_connection_lock_synced_data(connection);
+        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: Lock hold from s_mqtt_client_shutdown 1", (void *)connection);
```
graebm, line 123: could do something like this to keep your print statements, but not flood everyone
```
#define DEBUG_DEADLOCKS 0
#ifdef DEBUG_DEADLOCKS
#define LOGF_DEADLOCK_TRACE ....
```

source/client.c
```diff
@@ -116,6 +120,19 @@ static void s_mqtt_client_shutdown(
...
+        /* Move all the ongoing requests to the pending requests list, because the response they are waiting for will
+         * never arrives. Sad. But, we will retry. */
+        while (!aws_linked_list_empty(&connection->thread_data.ongoing_requests_list)) {
+            struct aws_linked_list_node *node =
+                aws_linked_list_pop_front(&connection->thread_data.ongoing_requests_list);
+            struct aws_mqtt_request *request = AWS_CONTAINER_OF(node, struct aws_mqtt_request, list_node);
+            aws_linked_list_push_back(&connection->synced_data.pending_requests_list, &request->list_node);
+        }
```
graebm, line 131: `aws_linked_list_move_all_back()`

source/client.c
```diff
@@ -1460,6 +1469,8 @@ int aws_mqtt_client_connection_disconnect(
 
     { /* BEGIN CRITICAL SECTION */
         mqtt_connection_lock_synced_data(connection);
```
graebm, line 1471: an idea, maybe bad:
have this be macro so you can use `__FUNCTION__` when printing and see where it's being called from

source/client.c
graebm: the behavior here is actually ok, but explain that aws_channel_slot_send_message() can only fail if the channel is already in the middle of shutting down

source/client.c
```diff
@@ -2484,7 +2501,7 @@ static enum aws_mqtt_client_request_state s_publish_send(uint16_t packet_id, boo
         }
 
         if (aws_channel_slot_send_message(task_arg->connection->slot, message, AWS_CHANNEL_DIR_WRITE)) {
-
+            /* TODO: NOT really agree on retry for this failure */
             aws_mem_release(message->allocator, message);
             return AWS_MQTT_CLIENT_REQUEST_ERROR;
         }
```
graebm, line 2507: this can fail if the channel is shutting down
it should probably do like subscribe and just pretend things are ok

source/client.c
```diff
@@ -2548,7 +2565,8 @@ uint16_t aws_mqtt_client_connection_publish(
     arg->on_complete = on_complete;
     arg->userdata = userdata;
 
-    uint16_t packet_id = mqtt_create_request(connection, &s_publish_send, arg, &s_publish_complete, arg);
+    uint16_t packet_id = mqtt_create_request(
+        connection, &s_publish_send, arg, &s_publish_complete, arg, qos == AWS_MQTT_QOS_AT_MOST_ONCE);
```
graebm, line 2569: trivial: readability:
```
bool retry = qos == AWS_MQTT_QOS_AT_MOST_ONCE;
uint16_t packet_id = mqtt_create_request(...., retry);
```

source/client_channel_handler.c
```diff
@@ -620,198 +623,98 @@ struct aws_io_message *mqtt_get_message_for_packet(
...
+                mqtt_connection_unlock_synced_data(connection);
+            } /* END CRITICAL SECTION */
+        } else {
+            AWS_LOGF_DEBUG(
+                AWS_LS_MQTT_CLIENT,
+                "static: task id %p, was canceled due to the channel shutting down. Request for packet id "
+                "%" PRIu16 ". will NOT be retried, will be cancelled",
+                (void *)task,
+                request->packet_id);
+            /* Fire the callback and clean up the memory, as the connection get destoried. */
```
graebm, line 655: trivial: destoried -> destroyed (search and replace)

graebm (approved):
fix & ship

source/client.c
TingDaoK: I chose to remove all the trace log instead😬

PR #156, opened 2020-10-06
fix deadlock

TingDaoK:
*Issue #, if available:*

- `aws_channel_shutdown` will cancel all the scheduled task, which may acquire the lock, and lead to a dead lock

*Description of changes:*

- don't call `aws_channel_shutdown` with lock hold

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -125,6 +125,7 @@ static void s_mqtt_client_shutdown(
             case AWS_MQTT_CLIENT_STATE_DISCONNECTING:
                 /* disconnect requested by user */
                 /* Successfully shutdown, so clear the outstanding requests */
+                /* TODO: this also invokes user callback, which is extremely dangerous */
                 aws_hash_table_clear(&connection->synced_data.outstanding_requests_table);
                 connection->synced_data.state = AWS_MQTT_CLIENT_STATE_DISCONNECTED;
                 break;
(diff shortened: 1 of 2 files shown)


tests/connection_state_test.c
```diff
@@ -1622,8 +1622,6 @@ static int s_test_mqtt_connection_disconnect_while_reconnecting(struct aws_alloc
     aws_channel_shutdown(state_test_data->server_channel, AWS_OP_SUCCESS);
     s_wait_for_interrupt_to_complete(state_test_data);
 
-    state_test_data->server_disconnect_completed = false;
```
TingDaoK, line 1625: this test will never reconnect, it's not needed to reset this Boolean.

source/client.c
```diff
@@ -125,6 +125,7 @@ static void s_mqtt_client_shutdown(
             case AWS_MQTT_CLIENT_STATE_DISCONNECTING:
                 /* disconnect requested by user */
                 /* Successfully shutdown, so clear the outstanding requests */
+                /* TODO: this also invokes user callback, which is extremely dangerous */
```
bretambrose, line 128: Honestly I think we should ban hash tables that do anything other than dead-simple deallocation of their key-value members in these callbacks.

bretambrose (approved):

source/client.c
TingDaoK: yeah, agreed. I'll try to improve this.

TingDaoK: removed from the other PR, https://github.com/awslabs/aws-c-mqtt/pull/154/commits/22272ebc962874844d24e049484d7297269117d7

PR #157, opened 2020-10-09
respect the clean session without calling disconnect

TingDaoK:
*Issue #, if available:*

- When the clean session is true, the retry policy should be different

*Description of changes:*

- When the connection lost for a clean session connection, all the uncompleted request will fail instead of retry
- When the new connection is a connection with clean session true, the previous session will be discarded.

*Note*
- From the [Spec 3.1.2.4](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718030) Session here for client only includes:
   - QoS 1 and QoS 2 messages which have been sent to the Server, but have not been completely acknowledged.
   - QoS 2 messages which have been received from the Server, but have not been completely acknowledged.

- In the event that the value of Session Present received by the Client from the Server is not as expected, the Client can choose whether to proceed with the Session or to disconnect. [Spec 3.2.2.2](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718030). We chose to proceed with the Session instead of disconnecting here.


source/client_channel_handler.c
```diff
@@ -111,25 +122,52 @@ static int s_packet_handler_connack(
...
                 struct aws_mqtt_request *request = AWS_CONTAINER_OF(current, struct aws_mqtt_request, list_node);
-                AWS_LOGF_TRACE(
-                    AWS_LS_MQTT_CLIENT,
-                    "id=%p: processing offline request %" PRIu16,
-                    (void *)connection,
-                    request->packet_id);
-                aws_channel_schedule_task_now(connection->slot->channel, &request->outgoing_task);
-
+                if (connection->clean_session) {
+                    /* discard the previous session, invoke the complete callbacks first */
```
graebm, line 134: We probably want to cancel existing requests when the user calls connect(), instead of when the connack comes back.

That makes it a little more clear that old requests are being cancelled because of the USER'S request for a clean session.

Also, it's a little confusing to let someone call connect(), publish(A), publish(B), and then A and B fail when the connection finally succeeds

source/client_channel_handler.c
```diff
@@ -797,12 +803,25 @@ uint16_t mqtt_create_request(
...
-            next_request->packet_id);
-        aws_channel_schedule_task_now(connection->slot->channel, &next_request->outgoing_task);
+        /* It's possible that after we release the lock, the slot is gone. So, let's check the slot. */
+        if (connection->slot && connection->slot->channel) {
+            AWS_LOGF_TRACE(
+                AWS_LS_MQTT_CLIENT,
+                "id=%p: Currently not in the event-loop thread, scheduling a task to send message id %" PRIu16 ".",
+                (void *)connection,
+                next_request->packet_id);
+            aws_channel_schedule_task_now(connection->slot->channel, &next_request->outgoing_task);
```
TingDaoK, line 813: it probably will not solve the root cause... but we cannot put a lock around the `aws_channel_schedule_task_now` function...

source/client.c
```diff
@@ -131,21 +130,15 @@ static void s_mqtt_client_shutdown(
                 "id=%p: Discard ongoing requests and pending requests when a clean session connection lost.",
                 (void *)connection);
             aws_linked_list_swap_contents(&connection->synced_data.pending_requests_list, &requests);
+            aws_linked_list_move_all_back(&requests, &connection->thread_data.ongoing_requests_list);
```
graebm: trivial: use move_all_back() in both places

also, add ongoing before pending so we can cancel in order

graebm: trivial: rename `requests` to `cancelling_requests`

source/client.c
```diff
@@ -274,6 +271,8 @@ static void s_mqtt_client_shutdown(
                     "id=%p: Caller requested disconnect from on_interrupted callback, aborting reconnect",
                     (void *)connection);
                 MQTT_CLIENT_CALL_CALLBACK(connection, on_disconnect);
+                /* Everything is done from the eventloop on the connection, leave the ownsership fully to user. */
+                aws_mqtt_client_connection_release(connection);
```
graebm, line 275: It's scary that this VERY IMPORTANT function is called from mulitple places in this very long and crazy function.

Maybe have 1 set_state() function, and that's where we acquire a reference when exiting DISCONNECTED state, or release the reference when entering DISCONNECTED state?

source/client.c
```diff
@@ -1431,7 +1430,57 @@ int aws_mqtt_client_connection_connect(
          * so we don't need to worry about it here*/
         goto error;
     }
+    /* Call will not fail after this. */
+    struct aws_linked_list requests;
+    aws_linked_list_init(&requests);
```
graebm, line 1435: trivial: rename `cancelling_requests`

source/client.c
```diff
@@ -1431,7 +1430,57 @@ int aws_mqtt_client_connection_connect(
...
+                    connection,
+                    request->packet_id,
+                    AWS_ERROR_MQTT_CANCELLED_FOR_CLEAN_SESSION,
+                    request->on_complete_ud);
+            }
+            current = current->next;
+        }
+        /* free the resource */
+        {
+            mqtt_connection_lock_synced_data(connection);
```
graebm, line 1471: trivial: forgot BEGIN CRITICAL SECTION

source/client_channel_handler.c
graebm: use aws_channel_acquire_hold() to keep it alive until you run this (don't forget to release_hold() after ;) )

graebm (approved):
fix & shup

PR #158, opened 2020-10-19
more information for unexpected hang up

TingDaoK:
*Issue #, if available:*

https://github.com/aws/aws-iot-device-sdk-js-v2/issues/101

*Description of changes:*

### DIFF
--- a/source/mqtt.c
+++ b/source/mqtt.c
@@ -137,7 +137,7 @@ bool aws_mqtt_is_valid_topic_filter(const struct aws_byte_cursor *topic_filter) 
                 "Library built without MQTT_WITH_WEBSOCKETS option."),
             AWS_DEFINE_ERROR_INFO_MQTT(
                 AWS_ERROR_MQTT_UNEXPECTED_HANGUP,
-                "The connection was closed unexpectedly."),
+                "The connection was closed unexpectedly. Internet lost. Or the server/broker hang up for reasons like permession, check the log from server/broker for the detail"),
             AWS_DEFINE_ERROR_INFO_MQTT(
                 AWS_ERROR_MQTT_CONNECTION_SHUTDOWN,
                 "MQTT operation interrupted by connection shutdown."),


source/mqtt.c
```diff
@@ -137,7 +137,7 @@ bool aws_mqtt_is_valid_topic_filter(const struct aws_byte_cursor *topic_filter)
                 "Library built without MQTT_WITH_WEBSOCKETS option."),
             AWS_DEFINE_ERROR_INFO_MQTT(
                 AWS_ERROR_MQTT_UNEXPECTED_HANGUP,
-                "The connection was closed unexpectedly."),
+                "The connection was closed unexpectedly. Internet lost. Or the server/broker hang up for reasons like permession, check the log from server/broker for the detail"),
```
bretambrose, line 140: Wording aside, I'm not sure this is what we want to do here; I don't think we should speculatively throw out a set of possible reasons this might happen.

justinboswell: +1. `The connection was closed unexpectedly` is sufficient

TingDaoK: hmmm, but several customers has provided feedback like when they have denied by the server because of permission issue, they may have no clue about what's happening... So, I feel like we should provide them some hint somewhere.

justinboswell: We don't have to on the client side. This is where they need to use cloudwatch to see why they were hung up on. The information simply isn't available to us on the client side to reason about, and burying a string in here of possibilities is guaranteed to go out of date.

https://docs.aws.amazon.com/iot/latest/developerguide/cloud-watch-logs.html

snobu: Then the hardcoded string should say "Go check your CloudWatch logs" and maybe even form the URL to it if you can. Not sure i understand how `The connection was closed unexpectedly` is sufficient in this case since there's absolutely no hint of where to go next from here, unless you intentionally want people to give up and go to competition, or this SDK is only for "AWS pros", so if you're just following the README you're going to have a bad time anyway.

TingDaoK: @snobu. We will add the information about unexpected hung up to the README as a trouble shooting page, or things like that. It's just too many reason for the unexpected hung up can happen to be included at the error message.

TingDaoK:
Need to add the information about unexpected hung up to the README as a trouble shooting page

PR #159, opened 2020-10-22
New Connect Functions

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- Rebase https://github.com/awslabs/aws-c-mqtt/pull/152 to the [retry of requests](https://github.com/awslabs/aws-c-mqtt/pull/154)
- Depercated original connect function and introduce two different connect functions, try connect and persistant connect.
- Depercated connected_complete callback
- Depercated on_interrupt/on_resume callback, introduce on_connected/on_disconnected callback with the event-handler.
- For on_connection/on_disconnected, they will be invoked when the connection connects/losts. The first connection will be same as the following, which has different behavior than the previous version


TingDaoK:
let's keep this PR, we will come back later.

PR #171, opened 2021-02-01
Test retain flag

graebm:
Split single test into 2 to ensure we're sending publish flags correctly


tests/packet_encoding_test.c
```diff
@@ -288,22 +288,61 @@ PACKET_TEST(CONNACK, connack, &s_test_connack_init, NULL, NULL)
...
+        aws_byte_cursor_from_array(s_payload, PAYLOAD_LEN)));
+
+    /* Init buffer */
+    /* clang-format off */
+    aws_byte_buf_write_u8(
+        &fixture->buffer,
+        (AWS_MQTT_PACKET_PUBLISH << 4) /* Packet type bits 7-4 */
+        | (1 << 3) /* DUP bit 3 */
+        | (AWS_MQTT_QOS_AT_MOST_ONCE << 1) /* QoS bits 2-1 */
+        | (0 << 0) /* RETAIN bit 0 */);
```
justinboswell, line 310: Shouldn't this be 1 << 0?

TingDaoK: retain is set to false in this test, so it's correct to be zero here.

TingDaoK (approved):
LGTM

tests/packet_encoding_test.c
graebm: it's correct, but yeah, bonkers looking. I'll remove the useless shift

PR #172, opened 2021-02-02
add args for dup, qos, retain to publish-received callback

graebm:
Don't hide anything from the user that they may wish to know.


include/aws/mqtt/client.h
```diff
@@ -90,11 +90,24 @@ typedef void(aws_mqtt_suback_fn)(
...
+ * \param[in] dup           DUP flag. If true, this might be re-delivery of an earlier attempt to send the message.
+ * \param[in] qos           Quality of Service used to deliver the message.
+ * \param[in] retain        Retain flag. If true, the message was sent as a result of a new subscription being
+ *                          made by the client.
+ */
 typedef void(aws_mqtt_client_publish_received_fn)(
     struct aws_mqtt_client_connection *connection,
     const struct aws_byte_cursor *topic,
     const struct aws_byte_cursor *payload,
+    bool dup,
```
justinboswell, line 108: Should this be a structure that we can update later? Won't MQTT5 add to this?

If we're going to break this, let's break it once.

graebm: I like your thinking, but there's no easy futureproofing of this API for MQTT5. MQTT5 has stuff like "the client could reject a PUBLISH from the server with a specific reason-code" that just can't be incorporated with what we have now. If we support MQTT5 in the future, we should make a distinct MQTT5 API, and have an implementation of this API that calls into it.

This is the one callback we know was missing information, now it has it all. It seems weird to have this one C callback take a struct while the others all take args. And it doesn't seem worth the effort at this point to switch ALL the callbacks to take structs.

justinboswell: :+1:

justinboswell (approved):

PR #175, opened 2021-02-11
Add more tests for CONNECT packet

graebm:
- test CONNECT with will
- test CONNECT with username/password


TingDaoK (approved):
It may also be good to add the similar tests to the connection_state_test.py, if it's not too hard.

PR #179, opened 2021-03-11
add timeout for publish

TingDaoK:
*Issue #, if available:*

- In case server side treat some publish package as not received, and not response to client, introducing a configurable timeout.
- same thing will happen to unsubscribe as well.

*Description of changes:*

- Timeout the publish QoS>0. When timeout happens, client will assume server never receives the publish and will not have any retry on the publish, just fail the publish.
- Same timeout for unsubscribe

*TODO:*
- Some optional retry strategy can help reduce the complexity for the users who want to resend the packet before timing out.


include/aws/mqtt/client.h
```diff
@@ -215,6 +212,7 @@ struct aws_mqtt_connection_options {
     struct aws_byte_cursor client_id;
     uint16_t keep_alive_time_secs;
     uint32_t ping_timeout_ms;
+    uint16_t publish_timeout_secs;
```
MikeDombo, line 215: can we use milliseconds instead?

TingDaoK: Sure, I just don't want to the timeout to be too short.

graebm (approved):

PR #189, opened 2021-04-13
Fix unsafe hash-table accesses

graebm:
Fact: It's not safe to mutate aws_hash_table while holding an aws_hash_element.
Fact: `outstanding_requests_table` is accessed by multiple threads. We always hold a lock while accessing it.

Bug: There were several places where we release the lock, but continue to hold an aws_hash_element we got from a find() call. Once the lock is released another thread can mutate the `outstanding_requests_table`, leading to undefined behavior when the aws_hash_element is accessed again.


source/client_channel_handler.c
```diff
@@ -819,27 +822,53 @@ uint16_t mqtt_create_request(
...
         }
+        /**
+         * Find a free packet ID.
+         * QoS 0 PUBLISH packets don't actually need an ID on the wire,
+         * but we assign them internally anyway just so everything has a unique ID.
+         *
+         * Yes, this is an O(N) search.
+         * We remember the last ID we assigned, so it's O(1) in the common case.
+         * But it's theoretically possible to reach O(N) where N is just above 64000
+         * if the user is letting a ton of un-ack'd messages queue up
```
graebm, line 833: Previously, we'd start searching from 0 each time for a free ID. This made things noticeably slow when the buffer was getting full. This makes it way faster. If anyone can think of a clever solution that isn't O(N) we can implement it in a follow-up

source/client_channel_handler.c
```diff
@@ -887,21 +916,38 @@ uint16_t mqtt_create_request(
...
         aws_hash_table_find(&connection->synced_data.outstanding_requests_table, &packet_id, &elem);
+        if (elem != NULL) {
+            found_request = true;
+
+            struct aws_mqtt_request *request = elem->value;
+            on_complete = request->on_complete;
+            on_complete_ud = request->on_complete_ud;
+
+            /* clean up request resources */
+            aws_hash_table_remove_element(&connection->synced_data.outstanding_requests_table, elem);
```
graebm, line 942: previously, we'd release the lock, invoke the completion callback, re-acquire the lock and THEN try to remove the aws_hash_element. If the hash-table was mutated in the meantime we'd be in a bad place.

Now we grab the lock, free everything, release the lock, then invoke the completion callbacks

bretambrose (approved):

source/client_channel_handler.c
justinboswell: As long as we're using a hash table, it's going to be tough to make this clever, other than to maintain a linear array of freed ids and use that once the table has been saturated once.

PR #191, opened 2021-04-13
First draft of elastipubsub

bretambrose:


source/client.c
```diff
@@ -2501,9 +2540,11 @@ static void s_unsubscribe_complete(
     struct unsubscribe_task_arg *task_arg = userdata;
 
     AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "id=%p: Unsubscribe %" PRIu16 " complete", (void *)connection, packet_id);
-    if (task_arg->timeout_arg) {
-        task_arg->timeout_arg->cancelled = true;
+    if (task_arg->timeout_wrapper.timeout_task_arg) {
```
graebm, line 2543: document this madness
`/* NULL out pointers so that, when timeout task runs, it knows that the request is already complete */`

source/client.c
```diff
@@ -2704,7 +2761,8 @@ static enum aws_mqtt_client_request_state s_publish_send(uint16_t packet_id, boo
         if (!timeout_task_arg) {
             return AWS_MQTT_CLIENT_REQUEST_ERROR;
         }
-        task_arg->timeout_arg = timeout_task_arg;
+        task_arg->timeout_wrapper.timeout_task_arg = timeout_task_arg;
```
graebm, line 2764: does this make any sense? add a comment like this, but making sense:

/* The request and its timeout task point to each other. Whichever one finishes first will null out both pointers. Then when the second one runs, it will know that it wasn't first */

source/client.c
```diff
@@ -2723,9 +2781,12 @@ static void s_publish_complete(
     if (task_arg->on_complete) {
         task_arg->on_complete(connection, packet_id, error_code, task_arg->userdata);
     }
-    if (task_arg->timeout_arg) {
-        task_arg->timeout_arg->cancelled = true;
+
+    if (task_arg->timeout_wrapper.timeout_task_arg != NULL) {
```
graebm, line 2785: `/* NULL out pointers so that, when timeout task runs, it knows that the request is already complete */`

source/client.c
```diff
@@ -94,11 +99,16 @@ static void s_request_timeout(struct aws_channel_task *channel_task, void *arg,
...
 
     if (status == AWS_TASK_STATUS_RUN_READY) {
-        if (!timeout_task_arg->cancelled) {
+        if (timeout_task_arg->task_arg_wrapper != NULL) {
             mqtt_request_complete(connection, AWS_ERROR_MQTT_TIMEOUT, timeout_task_arg->packet_id);
         }
     }
 
+    if (timeout_task_arg->task_arg_wrapper != NULL) {
+        timeout_task_arg->task_arg_wrapper->timeout_task_arg = NULL;
```
graebm, line 108: `/* NULL out pointers so that, when completion function runs, it knows that the timeout task has already been freed */

source/client.c
```diff
@@ -2467,7 +2505,8 @@ static enum aws_mqtt_client_request_state s_unsubscribe_send(
         if (!timeout_task_arg) {
             return AWS_MQTT_CLIENT_REQUEST_ERROR;
         }
-        task_arg->timeout_arg = timeout_task_arg;
+        task_arg->timeout_wrapper.timeout_task_arg = timeout_task_arg;
```
graebm, line 2508: something like...

/* The request and its timeout task point to each other. Whichever one finishes first will null out both pointers. Then when the second one runs, it will know that it wasn't first */

source/client.c
```diff
@@ -2723,9 +2810,18 @@ static void s_publish_complete(
     if (task_arg->on_complete) {
         task_arg->on_complete(connection, packet_id, error_code, task_arg->userdata);
     }
-    if (task_arg->timeout_arg) {
```
TingDaoK, line 2726: My bad😵 Thanks for fixing it.

source/client.c
```diff
@@ -80,12 +80,23 @@ void mqtt_connection_set_state(
...
+    struct request_timeout_wrapper *task_arg_wrapper;
+};
+
+/*
+ * We want the timeout task to be able to destroy the forward reference from the operation's task arg structure
+ * to the timeout task.  But the operation task arg structures don't have any data structure in common.  So to allow
+ * the timeout to refer back to a zero-able forward pointer, we wrap a pointer to the timeout task and embed it
+ * in every operation's task arg that needs to create a timeout.
+ */
+struct request_timeout_wrapper {
```
TingDaoK, line 98: Clever solution!

TingDaoK (approved):
LGTM

CMakeLists.txt
```diff
@@ -109,4 +109,7 @@ install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake"
 include(CTest)
 if (BUILD_TESTING)
     add_subdirectory(tests)
+    if (NOT CMAKE_CROSSCOMPILING )
+        add_subdirectory(bin/elastipubsub)
```
graebm, line 113: Bugfix looks good, but don't have time to review the sample app

can we leave that out for now? or would it slow you down testing further stuff?

graebm (approved):

CMakeLists.txt
bretambrose: Let's leave in for now.

PR #193, opened 2021-05-04
Fixed packet leak when after invalid MQTT packet

justinboswell:
Happens when CONNACK is not an MQTT valid packet, but is (for instance) an HTTP 501. Don't ask how I found this.

### DIFF
--- a/source/client_channel_handler.c
+++ b/source/client_channel_handler.c
@@ -525,7 +525,7 @@ static int s_process_read_message(
         AWS_ZERO_STRUCT(connection->thread_data.pending_packet);
 
         if (result) {
-            return AWS_OP_ERR;
+            goto cleanup;
         }
     }
 
@@ -561,10 +561,9 @@ static int s_process_read_message(
                 }
 
                 aws_reset_error();
-                goto cleanup;
-            } else {
-                return AWS_OP_ERR;
             }
+            /* done processing this packet, no more we can do */
+            goto cleanup;
         }
 
         struct aws_byte_cursor packet_data =


source/client_channel_handler.c
```diff
@@ -561,10 +561,9 @@ static int s_process_read_message(
                 }
 
                 aws_reset_error();
-                goto cleanup;
-            } else {
-                return AWS_OP_ERR;
             }
+            /* done processing this packet, no more we can do */
+            goto cleanup;
```
graebm, line 566: I don't think this is quite right.
if result != AWS_OP_ERR but aws_last_error() != AWS_ERROR_SHORT_BUFFER
then there's something fatally wrong with the incoming packet and we need to kill the connection.

previously, we returned AWS_OP_ERR, which (I'm guessing) resulted in the other channel handler shutting down the channel.

You changed it so we just clean up the message and return AWS_OP_SUCCESS, which will keep the channel open and more bytes might keep flowing into the MQTT handler as if nothing bad ever happened.

graebm: Ownership-wise, the MQTT code actually looks correct.
It's making assumptions that returning AWS_OP_ERR will have the side effect of killing the channel, but it's otherwise correct.

Is the bug that the TLS channel handler isn't cleaning up the aws_io_message if its aws_channel_send_message() call fails?

graebm: As an aside, I've always wanted to change this API. The ownership is so tricky to get right.
I would change it like so:
- aws_channel_handler_send_message() ALWAYS took ownership of the message and it wasn't the responsibility of the caller OR the receiver to clean it up.
- if the read_handler() vtable function reported an error, the aws_channel_handler_send_message() would ALWAYS kill the channel, instead of handler's needed to do it themselves, or assume their adjacent handler would do that

bretambrose:
The decode and framing refactor should handle this correctly now that the packet buffer stays active the entire connection

PR #196, opened 2021-08-19
Migrate CI to ecr and consolidate jobs

bretambrose:


.github/workflows/ci.yml
```diff
@@ -24,92 +27,57 @@ jobs:
...
+          - clang-6
+          - clang-8
+          - clang-9
+          - clang-10
+          - clang-11
+          - gcc-4.8
+          - gcc-5
+          - gcc-6
+          - gcc-7
+          - gcc-8
```
justinboswell, line 53: I just realized, gcc is up to 10, and clang is up to 12, should we be adding that to these repos?

justinboswell (approved):

.github/workflows/ci.yml
bretambrose: Someday!

PR #198, opened 2021-09-15
Allow for null on_message callbacks when subscribing to a topic

bretambrose:
* Under certain circumstances consumers may just want to use the on-any-message callback mechanism and skip subscribe on-publish callbacks entirely.  Let's not crash in that case.

### DIFF
--- a/include/aws/mqtt/client.h
+++ b/include/aws/mqtt/client.h
@@ -393,6 +393,8 @@ int aws_mqtt_client_connection_set_on_any_publish_handler(
  * Once the connection is opened, on_connack will be called. Only called when connection is disconnected.
  *
  * \param[in] connection                The connection object
+ * \param[in] connection_options        Configuration information for the connection attempt
+ *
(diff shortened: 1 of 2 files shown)


TingDaoK (approved):
was going to suggest adding comments to the header file to indicate that callback is not required anymore. But turns out we are not documenting it's required before😞. And, we have done null checks for **SOME** of the callbacks. But, still, if we can document it out about what can be NULL, what cannot, that will be super helpful.

bretambrose:
> was going to suggest adding comments to the header file to indicate that callback is not required anymore. But turns out we are not documenting it's required beforedisappointed. And, we have done null checks for **SOME** of the callbacks. But, still, if we can document it out about what can be NULL, what cannot, that will be super helpful.

I added nullable annotations to parameters that can be null for all of the operational APIs.

TingDaoK:
> I added nullable annotations to parameters that can be null for all of the operational APIs.

Awesome! Thanks!

PR #199, opened 2021-09-21
Pin the mqtt connection before kicking off the connect rather than after

bretambrose:
Should address https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/292

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -1526,7 +1526,16 @@ int aws_mqtt_client_connection_connect(
         } /* END CRITICAL SECTION */
     }
 
+    /* Begin the connecting process, acquire the connection to keep it alive until we disconnected */
+    aws_mqtt_client_connection_acquire(connection);
+
     if (s_mqtt_client_connect(connection, connection_options->on_connection_complete, connection_options->user_data)) {
+        /*
+         * An error calling s_mqtt_client_connect should (must) be mutually exclusive with s_mqtt_client_shutdown().
+         * So it should be safe and correct to call release now to undo the pinning we did a few lines above.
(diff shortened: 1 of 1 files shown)


source/client.c
```diff
@@ -1526,7 +1526,16 @@ int aws_mqtt_client_connection_connect(
...
 
+    /* Begin the connecting process, acquire the connection to keep it alive until we disconnected */
+    aws_mqtt_client_connection_acquire(connection);
+
     if (s_mqtt_client_connect(connection, connection_options->on_connection_complete, connection_options->user_data)) {
+        /*
+         * An error calling s_mqtt_client_connect should (must) be mutually exclusive with s_mqtt_client_shutdown().
+         * So it should be safe and correct to call release now to undo the pinning we did a few lines above.
+         */
+        aws_mqtt_client_connection_release(connection);
```
graebm, line 1537: We continue accessing `connection` later in this function

if we're worried about `connection` being cleaned up, shouldn't this `release()` call come after the the last time we access the variable?

TingDaoK (approved):

source/client.c
bretambrose: No, we enter this function with a ref count of 1, then the acquire above sends it to 2, now we need to undo the acquire above because our synchronous failure means the channel shutdown will not be called (that's where the matching release is for the pinning acquire),

graebm: ohhhhhhhhhhhhhh I'm dumb and didn't read the comment

graebm (approved):

PR #203, opened 2021-10-08
Support configurable include install dir

r-burns:
This is useful for split-output packaging, where the
development headers are separate from the runtime libraries.

*Description of changes:*

Installation directory for c header includes can be specified with the standard CMAKE_INSTALL_INCLUDEDIR variable from GNUInstallDirs. If unspecified, it defaults to `include` so is fully backward compatible.


graebm:
Fixed by: https://github.com/awslabs/aws-c-mqtt/pull/379

Issue #204, opened 2021-10-27
Allow receiving messages with empty payload


r7vme:
Hello!

Currently receiving packages with empty payload is [not possible](https://github.com/awslabs/aws-c-mqtt/blob/v0.7.8/source/packets.c#L587-L589).

MQTT 3.1 spec says that [payload is optional for PUBLISH packages](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718026).

Could such packages be allowed?

I can confirm publishing such packages is possible, but not subscribing.

---

###  Why empty messages are needed?

I'm using protocol buffers (proto3) and there is a possibility that serialized payload will be zero. For example, when all fields have [default values](https://developers.google.com/protocol-buffers/docs/proto3#default).

bretambrose:
This should be fixed in https://github.com/aws/aws-iot-device-sdk-cpp-v2/releases/tag/v1.14.7 and https://github.com/awslabs/aws-crt-cpp/releases/tag/v0.17.6

PR #206, opened 2021-11-10
Packet payload fixes

bretambrose:
Fixes for:
* https://github.com/awslabs/aws-c-mqtt/issues/204
* https://github.com/aws/aws-iot-device-sdk-java-v2/issues/188

As a next step, all of the CRTs should include empty will/payload tests as well.


TingDaoK (approved):

TingDaoK (approved):
Maybe, we could also add tests for decoding, like the fake server can send an empty payload

graebm (approved):

bretambrose:
> Maybe, we could also add tests for decoding, like the fake server can send an empty payload

The packet encoding tests do a decode and then compare at the end as well, so the empty publish tests are actually checking a decode of an empty payload.

PR #211, opened 2022-05-02
fixed endpoint not being read in arguments

sbSteveK:
small bug fix in elastipubsub

*Description of changes:*

endpoint argument was not being parsed correctly and is now being read and stored

### DIFF
--- a/bin/elastipubsub/main.c
+++ b/bin/elastipubsub/main.c
@@ -63,8 +63,8 @@ struct app_ctx {
 
 static void s_usage(int exit_code) {
 
-    fprintf(stderr, "usage: elastipubsub [options] endpoint\n");
-    fprintf(stderr, " endpoint: url to connect to\n");
+    fprintf(stderr, "usage: elastipubsub [options] --endpoint\n");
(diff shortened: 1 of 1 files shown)


bin/elastipubsub/main.c
```diff
@@ -158,24 +159,26 @@ static void s_parse_options(int argc, char **argv, struct app_ctx *ctx) {
             case 'h':
                 s_usage(0);
                 break;
+            case 0x02: {
```
TwistedTwigleg, line 162: Super minor question: Should we add a comment saying something like "start of text" or "stx" here? I found it a little confusing initially since the other switch statements are comparing against characters, but then once I searched up what `0x02` is mapped to it made sense, but I'm wondering if perhaps it would be best to add a quick comment just so at a glance it is more obvious what it is looking for.

TwistedTwigleg (approved):
👍

bin/elastipubsub/main.c
sbSteveK: Explicitly added the endpoint argument and instructions

TwistedTwigleg (approved):
Awesome! Looks great 👍

PR #214, opened 2022-06-08
Fix resubscribe segfault when there are no topics

### LINKED ISSUE #213: Segfault when resubscribing with no subscriptions
I have run into a segfault when resubscribing a connection that had no subscriptions to begin with.

I've debugged the issue as follows. `s_resubscribe_complete` is called even when resubscribing with no prior subscriptions due to the following data flow. `s_resubscribe_send` includes a check for not having any subscriptions:
```c
size_t sub_count = aws_mqtt_topic_tree_get_sub_count(&task_arg->connection->thread_data.subscriptions);
if (sub_count == 0) {
    AWS_LOGF_TRACE(
        AWS_LS_MQTT_CLIENT,
        "id=%p: Not subscribed to any topics. Resubscribe is unnecessary, no packet will be sent.",
        (void *)task_arg->connection);
    return AWS_MQTT_CLIENT_REQUEST_COMPLETE;
}
```
Normally, `s_resubscribe_send` allocates and initializes `task_arg->topics`. The segfault is caused by `task_arg->topics` being left uninitialized due to `s_resubscribe_send` exiting early from the above check.
According to the docs, `on_complete will be called once the request completed, either either in success or error`. Therefore, with `s_resubscribe_send` finished, `s_resubscribe_complete` runs. However, `s_resubscribe_send` never set `task_arg->topics` because it exited early. It is still null from the initial allocation inside `aws_mqtt_resubscribe_existing_topics`. When `s_resubscribe_complete` runs, it attempts to access `task_arg->topics` like so `aws_array_list_get_at(&task_arg->topics, &topic, 0)`. This causes a segfault.

I think I already have a decent understanding of this problem, so I can submit a patch for this, but I'd first like to know what ways in which the maintainers are open to solving this. I propose that `s_resubscribe_send` initializes the `topics` list as an empty list. We can then have `s_resubscribe_complete` check the size of this list before accessing its elements. An empty list would cause it to skip interacting with the list; it would then complete its cleanup steps and exit.

Octogonapus:
*Issue #, if available:*
Fixes #213.

*Description of changes:*
This PR fixes the segfault in #213 by always initializing the `task_arg->topics` list inside `s_resubscribe_send`.
When `s_resubscribe_complete` runs, it can then safely check the length of the list and exit if there are no topics.

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -2279,16 +2279,17 @@ static enum aws_mqtt_client_request_state s_resubscribe_send(
     bool initing_packet = task_arg->subscribe.fixed_header.packet_type == 0;
     struct aws_io_message *message = NULL;
 
-    size_t sub_count = aws_mqtt_topic_tree_get_sub_count(&task_arg->connection->thread_data.subscriptions);
+    const size_t sub_count = aws_mqtt_topic_tree_get_sub_count(&task_arg->connection->thread_data.subscriptions);
+    /* Init the topics list even if there are no topics because the s_resubscribe_complete callback will always run. */
+    if (aws_array_list_init_dynamic(&task_arg->topics, task_arg->connection->allocator, sub_count, sizeof(void *))) {
+        goto handle_error;
+    }
     if (sub_count == 0) {
         AWS_LOGF_TRACE(
             AWS_LS_MQTT_CLIENT,
             "id=%p: Not subscribed to any topics. Resubscribe is unnecessary, no packet will be sent.",
             (void *)task_arg->connection);
         return AWS_MQTT_CLIENT_REQUEST_COMPLETE;
-    }
-    if (aws_array_list_init_dynamic(&task_arg->topics, task_arg->connection->allocator, sub_count, sizeof(void *))) {
-        goto handle_error;
     }
     aws_mqtt_topic_tree_iterate(&task_arg->connection->thread_data.subscriptions, s_reconnect_resub_iterator, task_arg);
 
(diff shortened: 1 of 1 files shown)


bretambrose:
Zeroed array lists can be safely queried for length.  Just eyeballing everything, it looks like a simpler fix would be to use aws_mem_calloc instead of aws_mem_acquire (a change that must have been missed years ago when we switched to calloc-ing everything) here: https://github.com/awslabs/aws-c-mqtt/blob/main/source/client.c#L2408

Can you see if that fixes your issue and if so, we can pull in an updated PR.

Octogonapus:
I am not sure they can be safely queried for length because of the `aws_array_list_is_valid` precondition. I ran my code on the current state of this PR and got this:
```c
Fatal error condition occurred in /workspace/x86_64-linux-gnu-libgfortran3-cxx11/destdir/include/aws/common/array_list.inl:339: aws_array_list_is_valid(list)

(rr) bt
#0  raise (sig=<optimized out>) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007fa3a2d542e5 in aws_debug_break () at /workspace/srcdir/aws-c-common/source/posix/system_info.c:156
#2  0x00007fa3a2d36158 in aws_fatal_assert (cond_str=0x7fa3a30c586e "aws_array_list_is_valid(list)", 
    file=0x7fa3a30c57b8 "/workspace/x86_64-linux-gnu-libgfortran3-cxx11/destdir/include/aws/common/array_list.inl", line=339) at /workspace/srcdir/aws-c-common/source/assert.c:14
#3  0x00007fa3a2de9ed7 in aws_array_list_length (list=0x298a748) at /workspace/x86_64-linux-gnu-libgfortran3-cxx11/destdir/include/aws/common/array_list.inl:339
#4  0x00007fa3a2df1841 in s_resubscribe_complete (connection=0x28faa60, packet_id=2, error_code=0, userdata=0x298a740) at /workspace/srcdir/aws-c-mqtt/source/client.c:2361
```

bretambrose:
Well that's annoying.  Let's go with the original approach then.  I'll create an authorized PR, might make some minor changes to minimize unnecessary changes.

bretambrose:
PR tested and merged in https://github.com/awslabs/aws-c-mqtt/pull/215

PR #228, opened 2022-11-23
Fix zip and update canary time

TwistedTwigleg:
*Description of changes:*

Everything worked fine with the Canary except uploading the zip. This was due to using a `zsh` command rather than a `bash` one. This new command for getting only readable files should, I think, work as expected.

Also sets the 24/7 Canary to run for 7 hours instead of 2 now that it passed the initial test.

_______


TwistedTwigleg:

PR #232, opened 2022-11-26
 Decoder logging and several canary improvements

TwistedTwigleg:
*Description of changes:*

Makes the following changes to the MQTT code:
* Adds additional logging around decode errors for MQTT5 for easier debugging on what caused a decoder error.

Makes the following canary adjustments:
* Adjust canary to report OS level metrics rather than process level with a PID
  * The reason for this is that in certain environments, like `mvn`, tracking a PID is not sufficient and will result in the `mvn` process being tracked and not the thing that is actually running, even if processing is happening. This will happen for any process we want to monitor where the process spawns another process adjacent to it.
  * Another reason for this is that, in Codebuild, the only things running are going to be the Canary anyway. While we will grab things like the Python Canary Wrapper itself, these things should be **consistent** across Codebuild runs.
  * **In the future** we want to use a better solution so we can for sure just track the PID(s) we want to track and nothing else, but in the short term, we can just track OS metrics to still get usable metrics in the meantime.
* `stdout` is no longer captured, causing issues where sometimes the return code would not be noticed and therefore the application monitor would wait forever for an application that already finished.
* Capture `stdout` output by using a `tee` pipe to send the output to a file, and then read that file to add it to the canary log.
* `=====` is only printed on new runs when the canary restarts or finishes.
* Time stamps are now `Hour:Minute:Second` rather than `Hour-Minute-Second`.
* Metrics that track percentages have `%` in the logs

__________

### DIFF
--- a/codebuild/CanaryWrapper.py
+++ b/codebuild/CanaryWrapper.py
@@ -123,7 +123,8 @@ data_snapshot.register_metric(
     new_metric_unit="Percent",
     new_metric_alarm_threshold=70,
     new_metric_reports_to_skip=1,
-    new_metric_alarm_severity=5)
+    new_metric_alarm_severity=5,
+    is_percent=True)
 data_snapshot.register_metric(
     new_metric_name="total_memory_usage_value",
     new_metric_function=get_metric_total_memory_usage_value,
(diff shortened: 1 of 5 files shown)


bretambrose (approved):

TwistedTwigleg:

PR #238, opened 2022-12-08
TMP: Only print packet on error

TwistedTwigleg:
*Description of changes:*

To reduce the amount of data in the console, these adjustments will print the packet data only when an error is discovered. Additionally, all of the logs for printing packet data was moved from DEBUG to ERROR. With these changes, the amount of text in the logs when an error occurs should be a little easier to parse without taking huge amounts of data.

__________


bretambrose (approved):

TwistedTwigleg:

PR #241, opened 2022-12-12
More debugging for MQTT5 decoder

TwistedTwigleg:
*Description of changes:*

Adds even more debugging logs to the MQTT5 decoder, only applies for temporary debugging.
Now if an unknown/unimplemented packet type enum is sent, the enum value is printed as an error. Print statements were also added after running the three main decoder functions to see which had an error.

____________


bretambrose (approved):
This is fine, but at this point the right way forward is probably to support rolling of some kind in the logger.

TwistedTwigleg:

PR #252, opened 2023-01-23
General improvements to MQTT5 canary

TwistedTwigleg:
*Description of changes:*

Makes the following improvements to the MQTT5 canary:
* Allows running the MQTT5 canary forever if the seconds to run is set to `0`
* Allows setting the port for the Canary to use via `-p <port>` or `--port <port>`
* Adds an ACK timeout for the operations and callbacks for publish, subscribe, and unsubscribe that will print the error code should an error occur

_________________


TwistedTwigleg:

PR #254, opened 2023-01-23
Backed up socket fixes

TwistedTwigleg:
*Description of changes:*

Adjusts the MQTT311 client with the following fixes designed to help when the MQTT socket is congested with packet data:
* Adjusts the reconnect task to skip reconnecting if the MQTT connection is disconnected.
* Adjusts the ping task so that it will only send a ping if the last outbound socket write is longer than the duration of the ping timeout.
  * Prevents disconnects due to ping timeouts when the socket is full of data and the ping cannot get out in time.
  * This is within the MQTT specification, as it states  "*maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one Control Packet and the point it starts sending the next*".
    * Normally we send the ping irrespective of the data going out, but on congested sockets full of packet data, this can lead to the ping not getting out and the client disconnecting because it does not get a ping response. By tracking the last sent packet, we can ensure we stay within the specification while avoiding disconnects due to ping timeouts.

___________________


source/client_channel_handler.c
```diff
@@ -799,6 +842,14 @@ static void s_request_outgoing_task(struct aws_channel_task *task, void *arg, en
...
                     request->connection, request, AWS_MQTT_OSS_NONE);
 
+                /**
+                 * Cache the time of the write on the socket so long as the packet was not a ping packet.
+                 * We do this to avoid the PING packet overriding the ping interval
+                 */
+                if (request->packet_type != AWS_MQTT_PACKET_PINGREQ) {
+                    aws_high_res_clock_get_ticks(&connection->last_outbound_socket_write_time);
+                }
+
```
TwistedTwigleg, line 852: *Note* We can avoid needing to pass the packet type entirely by just checking the packet size and seeing if it is over `0`, as we assign a packet size to everything but the PING tasks currently.

However, I was not totally sure if this was the right approach because there is a possibility (though likely slim) that we have packets going out that we do not necessarily want to track the size of, but we do want to have reset the ping. Examples of this could be QoS2 ACK packets.

Either approach should work though, but I thought it would be better to be more targeted in the condition checking for the PING packet than not.

source/client.c
```diff
@@ -612,6 +612,36 @@ static void s_attempt_reconnect(struct aws_task *task, void *userdata, enum aws_
 
         mqtt_connection_lock_synced_data(connection);
 
+        /**
+         * Check the state and determine if this reconnect is happening in the middle of a disconnect! If it is, then
+         * abort the reconnect and finish the disconnect (as the disconnect is now deadlocked thanks to the reconnect)
+         */
+        if (connection->synced_data.state == AWS_MQTT_CLIENT_STATE_DISCONNECTING) {
```
TwistedTwigleg, line 619: I am not super confident in this code, though it does seem to work in all tests and fixes a deadlock I found in the `s_test_mqtt_connection_disconnect_while_reconnecting` test.

What seems to be the case in the test above is that the reconnect task happening in the middle of the disconnect causes a deadlock where the disconnect callback is never fired because the reconnect overrides it. If you do not complete the disconnect here, the test will hang either waiting for the disconnect to complete, or if you call ` MQTT_CLIENT_CALL_CALLBACK(connection, on_disconnect);` then it finishes the test but hangs waiting for memory to be cleaned up. Finishing the disconnect manually and ensuring everything is cleaned up seems to fix it in the tests I've run locally with CTEST.

Doing a straight cancel/abort of the reconnect task leads to this deadlock happening every time, while with this fix the deadlock never occurs.

However, I am not super confident that this is the best way to solve it...

source/client_channel_handler.c
TwistedTwigleg: If we remove adding the type along with the request, then we can trim the changes this PR makes without any loss of functionality currently.

TwistedTwigleg: I have changed my mind again. For now, we can just use the operation size and should we need to change it in the future, we can do so then. No reason to prematurely optimize and add quite a bit of code for a *maybe*.

source/client.c
TwistedTwigleg: **OKAY** I have looked further into this and was right in the general idea but wrong in what was causing it, why it was occurring, and how the fix solved it.

The situation is pretty simple: We have a disconnection that happens right in the middle of a reconnect. When this happens, we do not want to reconnect anymore.

The problem is that the disconnect function, `aws_mqtt_client_connection_disconnect`, will ONLY set the state of the MQTT client to `AWS_MQTT_CLIENT_STATE_DISCONNECTING` if it is in the middle of a reconnect because there is no connection `slot` set, essentially making all it does change the state.
- Previously, this was not an issue because in the reconnect task, we would ALWAYS try to reconnect and this would trigger a connection, spin up a new channel and `slot`, see the state is set to disconnecting, and then shut everything down. However, this is what we want to avoid as that is a decent amount of work that we immediately throw away.

So, the solution: We have to check in the reconnect task whether or not there is a `slot` and if there is NOT a slot, we finish the disconnect ourselves and fully clean up the connection as if the disconnection fully happened. If there IS a slot, then we just cancel like normal because the disconnect task will be properly scheduled and we can just let it clean itself up like normal.
- If we do not perform the disconnect ourselves and fully clean up the connection, then we get a DEADLOCK forever as the client does not A) have it's connection ref count reduced to 0 properly and B) it does not every invoke the disconnection customer callback, so any code the customer may have will wait forever for a disconnect that will never arrive.
- Note that I tried finishing the disconnect from the disconnect itself, but due to how the tasks are laid out, doing this always leads to segfaults.
- Annoyingly, we cannot "properly" disconnect from the disconnect function itself because we do not have a slot to schedule a task on and, even if we did, we'd be racing with the already scheduled reconnect task...
- If there IS a `slot` and the disconnect is called while reconnecting, then just canceling the reconnect without calling the disconnect callback will work fine, because the disconnect task will be properly scheduled on the `slot` in `aws_mqtt_client_connection_disconnect`.
- If the reconnect task happens BEFORE the disconnect that is also fine, it will just attempt to reconnect because the state will not be `AWS_MQTT_CLIENT_STATE_DISCONNECTING`.

This change should do everything we want and I'm decently confident now that it is the right solution. It should also not change the existing functionality for customers at all and should act as they expect.

source/client_channel_handler.c
```diff
@@ -36,21 +36,50 @@ static int s_packet_handler_default(
...
+    s_schedule_ping_get_outbound_delta_data(connection, &now, &outbound_delta, &ping_time_ns);
+
     AWS_LOGF_TRACE(
-        AWS_LS_MQTT_CLIENT, "id=%p: Scheduling PING. current timestamp is %" PRIu64, (void *)connection, now);
+        AWS_LS_MQTT_CLIENT, "id=%p: Scheduling PING task. current timestamp is %" PRIu64, (void *)connection, now);
 
-    uint64_t schedule_time =
-        now + aws_timestamp_convert(connection->keep_alive_time_secs, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+    uint64_t schedule_time = 0;
+    if (outbound_delta < ping_time_ns) {
```
bretambrose, line 74: This is kind of weird and I think it's related to the fact that you're skipping updating the last outbound write time on pings.  I haven't mentally walked through things in detail, but I feel like pings shouldn't be treated especially and in doing so, it will simplify these time calculations which are not what I'd expect.  

I would expect this function to work properly no matter when it's called, but as it stands that's  not the case.  It seems coupled to a hidden "this function only gets called right after I've sent a ping that didn't update the last write time"

I can't tell if it's correct or not, but I can tell it would be unnecessarily tough to maintain after the fact.

Actually it doesn't seem correct.  Here's the basic scenario I'm blocked on:

Assume 30 second keep alive

T(0) - Connect and schedule ping
T(29) - Publish something, last outbound is updated
T(30) - Ping task goes off.  s_on_time_to_ping gets called, but we skip sending it (else clause below) and just call s_schedule_ping.  In s_schedule_ping, outbound_delta is 1, so schedule time is 30 + 30 + 1 = 61 when it should be 59.  It only gets worse the closer the publish is in time to the previous ping/connect timepoint.

TwistedTwigleg: I fixed this. Now the PING packet is not treated specially and it works as expected. Now it marks the time regardless of whether it is a PING packet or another packet. I did have to add a slight delta check in the condition that checks to see if the time between the last write is less than the given keep alive, as a packet is not perfectly written exactly to the socket at the exact time (so we have to account for it). Now the code will account for 0.001 second delta accordingly.

I also fixed the reconnect timing so it should be as you mentioned, where it schedule at 59 rather than 61. I did some tests and I'm fairly confident that it functions like you said, where the schedule time will be 59 instead of 60 in the basic scenario given.

TwistedTwigleg: However, I probably should do additional testing and make a test that basically is the scenario you mentioned above (with shorter times) to ensure that it is correctly working.

TwistedTwigleg: Tests added! Now I am confident it is working correctly (even if in the second test I couldn't quite get the test suite to work how I wanted). I have checked the logs and confirmed that it should now work as expected in the basic scenario above.

source/client_channel_handler.c
```diff
@@ -36,21 +47,67 @@ static int s_packet_handler_default(
...
+    // We just need the delta, so biggest minus smallest. We do NOT want to overflow
+    if (aws_sub_u64_checked(now, connection->last_outbound_socket_write_time, &outbound_delta) != AWS_OP_SUCCESS) {
+        aws_sub_u64_checked(connection->last_outbound_socket_write_time, now, &outbound_delta);
+    }
+
+    uint64_t ping_time_ns =
+        aws_timestamp_convert(connection->keep_alive_time_secs, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+
+    // If there has not been a socket write yet, then the delta should be the scheduled ping timeout.
+    if (connection->last_outbound_socket_write_time == 0) {
```
xiazhvera, line 68: We can also do the zero check before `aws_sub_u64_checked` to optimize a little bit. Also, if we do apply the change, we can also reset `last_outbound_socket_write_time` to 0 in `aws_mqtt_client_connection_connect()`, so that we dont need to calculate the delta time when it comes to a new connection.

xiazhvera (commented):
I didn't approve the PR as I'm not confident about the fix with reconnection, though I could not think of a better solution. It would be better to have Bret to take a look. 
Other than that it looks good to me!

include/aws/mqtt/private/client_impl.h
```diff
@@ -127,6 +127,8 @@ struct aws_mqtt_request {
     enum aws_mqtt_operation_statistic_state_flags statistic_state_flags;
     /* The encoded size of the packet - used for operation statistics tracking */
     uint64_t packet_size;
+    /* The type of the packet */
+    enum aws_mqtt_packet_type packet_type;
```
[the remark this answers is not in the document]
TwistedTwigleg: Ah, yeah, that was a reminant of an older change where I stored the packet type per request but I forgot to remove it from the header... 😅 
Good catch! I removed it 👍

source/client_channel_handler.c
```diff
@@ -817,6 +848,9 @@ static void s_request_outgoing_task(struct aws_channel_task *task, void *arg, en
                 mqtt_connection_unlock_synced_data(connection);
             } /* END CRITICAL SECTION */
 
+            /* Since a request has complete, update the next ping time */
+            s_update_next_ping_time(connection);
```
bretambrose, line 852: Let's move this into the critical section above.  Update references channel/slot which could be side-affected under us if we're not in the event loop thread.

TwistedTwigleg:

PR #256, opened 2023-01-24
add utf8 validation check on user properties

sbSteveK:
Added utf8 validation check on properties for outbound packets. 
List of the validated properties:
- CONNECT
     - client_id
     - username
- DISCONNECT (outbound)
     - reason string
- PUBLISH
     - topic
     - payload (When payload format is set to UTF8)
     - response_topic
     - content_type
- UNSUBSCRIBE
     - topic filter
- SUBSCRIPTION
     - topic filter

The validation for inbound packets will be added in the future. Here is a list for inbound properties need covered in this PR for future reference:
- CONNACK
            - response information
            - client id
            - reason string
            - server reference
            - authentication method (currently not supported)
- PUBACK/PUBREC/PUBREL/PUBCOMP/SUBACK/UNSUBACK
            - reason string
            - user properties
- DISCONNECT (inbound)
            - server reference
            - reason string
            - user properties


bretambrose (changes requested):
If we're going to do outbound utf-8 validation, I would do it for everything appropriate.

bretambrose (changes requested):
Couple more suggestions:

1. List all of the newly-validated fields in the PR description for easy x-ref with the spec
1. The spec disallows certain codepoints that are still valid utf-8.  It also contains a section on the consequences of allowing those forbidden codepoints.  We need to look beyond just "valid utf-8" and consider the additional constraints in the spec.

source/v5/mqtt5_utils.c
```diff
@@ -537,3 +537,18 @@ bool aws_mqtt_is_topic_filter_shared_subscription(struct aws_byte_cursor topic_c
 
     return true;
 }
+
+// UTF-8 encoded string validation respect to [MQTT-1.5.3-2].
+bool aws_mqtt_utf8_validator(const uint32_t codepoint) {
+    // Control characters are not valid
+    if ((codepoint >= 0x0001 && codepoint <= 0x001F) || (codepoint >= 0x007F && codepoint <= 0x009F)) {
```
bretambrose, line 544: A good optimizer would probably make these changes for you, but let's be careful and explicit on this since it's a per-character inner loop:

1. Remove the ( == 0) check on 549 and change the first if to just <= 0x1F
2. Change the  && 0xFFFF compound check to ((codepoint & 0xFFFF) >= 0xFFFE)

So

```
if (codepoint <= 0x1F) {
   return false;
}

if (codepoint >= 0x7F && codepoint <= 0x9F) {
   return false;
}

if ((codepoint & 0xFFFF) >= 0xFFFE) {
   return false;
}

if (codepoint >= 0xFDD0 && codepoint <= 0xFDEF) {
   return false;
}

return true;
```

tests/v5/mqtt5_operation_and_storage_tests.c
```diff
@@ -161,24 +161,24 @@ static const struct aws_mqtt5_user_property s_user_properties[] = {
         .name =
             {
                 .ptr = (uint8_t *)s_user_prop1_name,
-                .len = AWS_ARRAY_SIZE(s_user_prop1_name),
+                .len = AWS_ARRAY_SIZE(s_user_prop1_name) - 1,
```
xiazhvera, line 164: -1 to remove the null terminator.

TwistedTwigleg (approved):

include/aws/mqtt/private/v5/mqtt5_utils.h
```diff
@@ -96,6 +97,12 @@ AWS_EXTERN_C_BEGIN
  */
 AWS_MQTT_API extern struct aws_byte_cursor g_aws_mqtt5_connect_protocol_cursor;
 
+/**
+ * utf8_decoder_options, used to setup the utf8_decoder
+ * The option will setup the callback to validate the utf8 string following mqtt5 specs
+ */
+AWS_MQTT_API extern struct aws_utf8_decoder_options g_aws_mqtt5_utf8_decoder_options;
```
bretambrose, line 104: Would rather see a utility function that keeps the options internal and hidden

bretambrose (approved):

PR #258, opened 2023-02-03
Add on_closed callback to MQTT311

TwistedTwigleg:
*Description of changes:*

Adds a `on_closed` callback that is called whenever a MQTT311 connection is fully closed. This is similar to the `on_disconnect` callback, but unlike that callback, this callback can be set once and will be called for every single successful, user-initiated disconnect, unlike the `on_disconnect` callback that is set each time you call `disconnect`.

This is for exposing this callback to the CRTs, which will allow the on_closed callback to be bubbled up akin to the V1 SDK API (reference).

Also adds tests to ensure the functionality works as expected and only is invoked on successful disconnects and not interrupts.

**Note:** If #254 is merged first, then this PR will need modification to call the `on_closed` callback when the disconnect is invoked via the reconnect, and vice versa should this PR be merged first.

____________


include/aws/mqtt/client.h
```diff
@@ -63,6 +63,9 @@ typedef void(aws_mqtt_client_on_connection_interrupted_fn)(
     int error_code,
     void *userdata);
 
+/* Called if the connection to the server is closed by user request */
+typedef void(aws_mqtt_client_on_connection_closed_fn)(struct aws_mqtt_client_connection *connection, void *userdata);
```
sbSteveK, line 67: Do we want to add a struct to this callback for future possible extension? We currently use an empty struct for that possibility in MQTT5 in callbacks like lifecycle events stop and attempting connect.

sbSteveK (approved):

TwistedTwigleg:

PR #269, opened 2023-04-06
Resolution override

bretambrose:

### DIFF
--- a/include/aws/mqtt/client.h
+++ b/include/aws/mqtt/client.h
@@ -387,6 +387,14 @@ int aws_mqtt_client_connection_set_http_proxy_options(
     struct aws_http_proxy_options *proxy_options);
 
 /**
+ * Set host resolution ooptions for the connection.
+ */
+AWS_MQTT_API
+int aws_mqtt_client_connection_set_host_resolution_options(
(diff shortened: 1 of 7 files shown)


include/aws/mqtt/client.h
```diff
@@ -386,6 +386,14 @@ int aws_mqtt_client_connection_set_http_proxy_options(
     struct aws_mqtt_client_connection *connection,
     struct aws_http_proxy_options *proxy_options);
 
+/**
+ * Set host resolution ooptions for the connection.
+ */
+AWS_MQTT_API
+int aws_mqtt_client_connection_set_host_resolution_options(
```
TwistedTwigleg, line 393: Later we'll need to bind these new options, correct? If so, will we put them in the MQTT5 client options?

TwistedTwigleg (approved):

include/aws/mqtt/client.h
TwistedTwigleg: MQTT5 should be fairly straight forward to add support for in the bindings, but I'm not 100% sure right off with MQTT311. Though that will be in the future, so we can cross that bridge when we get there.

PR #274, opened 2023-04-19
Mqtt311 shared subscription support

TwistedTwigleg:
*Description of changes:*

Checks subscribe and unsubscribe topic filters in MQTT311 to see if they start with `$share` and, if they do, then instead of assigning the callback in the topic tree to `$share/<group id>/<topic>`, it instead assigns it to `<topic>` so the callbacks will be invoked properly.

~~Still a work in progress! Currently untested at this point but testing is on-going.~~
Edit: Tested with a modified PubSub sample in the C++ SDK and confirmed it works as expected.

__________

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -174,12 +174,9 @@ static void s_init_statistics(struct aws_mqtt_connection_operation_statistics_im
 }
 
 static bool s_is_topic_shared_topic(struct aws_byte_cursor *input) {
-    struct aws_byte_cursor split_part;
-    AWS_ZERO_STRUCT(split_part);
-    if (aws_byte_cursor_next_split(input, '/', &split_part)) {
-        if (aws_byte_cursor_eq_c_str(input, "$share")) {
-            return true;
-        }
+    char *input_str = (char *)input->ptr;
+    if (strncmp("$share/", input_str, strlen("$share/")) == 0) {
+        return true;
     }
     return false;
 }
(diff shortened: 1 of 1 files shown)


source/client.c
```diff
@@ -173,6 +173,54 @@ static void s_init_statistics(struct aws_mqtt_connection_operation_statistics_im
...
+    }
+    // Make sure there is something in <topic>, properly detecting "$share/<group identifier>/" as invalid
+    if (split_part.len <= 0) {
+        AWS_LOGF_ERROR(
+            AWS_LS_MQTT_CLIENT,
+            "Cannot parse shared subscription topic: There is no topic in shared subscription topic filter");
+        return NULL;
+    }
+
+    struct aws_string *result_string = aws_string_new_from_c_str(input->allocator, (char *)split_part.ptr);
```
sbSteveK, line 220: Check to make sure if a topic has an `/` character in it, that we aren't only returning a partial topic. I think aws_byte_cursor_next_split isn't assigning `split_part` as the entire remainder of the `input_cursor` but just the next portion of it between the split characters.

source/client.c
```diff
@@ -173,6 +173,40 @@ static void s_init_statistics(struct aws_mqtt_connection_operation_statistics_im
...
+        return true;
+    }
+    return false;
+}
+
+static struct aws_string *s_get_normal_topic_from_shared_topic(struct aws_string *input) {
+    const char *input_char_str = aws_string_c_str(input);
+    int input_char_length = strlen(input_char_str);
+    int found_slashes = 0;
+    int split_position = 0;
```
sbSteveK, line 188: trivial:
When this is called, we're certain that the input starts with "$share" so we could start the check at index 7 and construct the return result at the first encountered '/' or fall from the while directly to the ERROR and return NULL.

source/client.c
```diff
@@ -173,6 +173,40 @@ static void s_init_statistics(struct aws_mqtt_connection_operation_statistics_im
...
+            if (found_slashes == 2) {
+                break;
+            }
+        }
+    }
+    // A properly formatted shared subscription topic will be "$share/<group id>/<topic>" and that means there needs
+    // to be 2 slashes at least. The loop above stops at 2, so if it is NOT 2 then the input was invalid
+    if (found_slashes == 2 && split_position + 1 <= input_char_length) {
+        char result_char[input_char_length - split_position];
+        strncpy(result_char, input_char_str + split_position + 1, input_char_length - split_position);
```
sbSteveK, line 202: trivial:
I feel like I've seen `sizeof(result_char)` style for `strncpy` to a specific char array. Unsure if that's a format we need to adhere to but just pointing it out because it's fresh in my head. This is also fine!

sbSteveK (approved):

source/client.c
```diff
@@ -173,6 +173,40 @@ static void s_init_statistics(struct aws_mqtt_connection_operation_statistics_im
...
+            found_slashes += 1;
+            if (found_slashes == 2) {
+                break;
+            }
+        }
+    }
+    // A properly formatted shared subscription topic will be "$share/<group id>/<topic>" and that means there needs
+    // to be 2 slashes at least. The loop above stops at 2, so if it is NOT 2 then the input was invalid
+    if (found_slashes == 2 && split_position + 1 <= input_char_length) {
+        char result_char[input_char_length - split_position];
```
sbSteveK, line 201: It looks like we need to make sure this value isn't 0? Windows warnings as error needs to be fixed but outside of that, looks good.

source/client.c
TwistedTwigleg: It looks like it is split. In `system_info.c` we use calculations like we do here, while in `socket_test.c` we use `sizeof`. It seems hit and miss. For now, I'm just going to leave it, as especially after making the change to make Windows happy, it is pretty straight forward without doing any additional calculations.
Good catch though! I didn't think to see if we had a pattern or not.

TwistedTwigleg:

PR #276, opened 2023-04-25
add a millisecond sleep to insure tests dont share socket endpoints

sbSteveK:
Tests that ran too close to one another were causing identical endpoints for local sockets. Added a millisecond sleep during endpoint generation. 

*Description of changes:*

### DIFF
--- a/tests/v3/connection_state_test.c
+++ b/tests/v3/connection_state_test.c
@@ -26,6 +26,7 @@
 
 static const int TEST_LOG_SUBJECT = 60000;
 static const int ONE_SEC = 1000000000;
+static const int ONE_MILLISECOND = 1000000
 // The value is extract from aws-c-mqtt/source/client.c
 static const int AWS_RESET_RECONNECT_BACKOFF_DELAY_SECONDS = 10;
 static const uint64_t RECONNECT_BACKOFF_DELAY_ERROR_MARGIN_NANO_SECONDS = 500000000;
@@ -243,6 +244,8 @@ static int s_setup_mqtt_server_fn(struct aws_allocator *allocator, void *ctx) {
     ASSERT_SUCCESS(aws_condition_variable_init(&state_test_data->cvar));
     ASSERT_SUCCESS(aws_mutex_init(&state_test_data->lock));
 
+    /* Sleep for one millisecond to insure timestamp doesn't repeat over tests */
+    aws_thread_current_sleep(ONE_MILLISECOND);
     uint64_t timestamp = 0;
     ASSERT_SUCCESS(aws_sys_clock_get_ticks(&timestamp));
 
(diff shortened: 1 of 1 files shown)


tests/v3/connection_state_test.c
```diff
@@ -26,8 +26,9 @@
 
 static const int TEST_LOG_SUBJECT = 60000;
 static const int ONE_SEC = 1000000000;
-// The value is extract from aws-c-mqtt/source/client.c
-static const int AWS_RESET_RECONNECT_BACKOFF_DELAY_SECONDS = 10;
+static const int ONE_MILLISECOND = 1000000
+    // The value is extract from aws-c-mqtt/source/client.c
+    static const int AWS_RESET_RECONNECT_BACKOFF_DELAY_SECONDS = 10;
```
TwistedTwigleg, line 31: Nitpick: Remove indent?

tests/v3/connection_state_test.c
```diff
@@ -243,14 +244,16 @@ static int s_setup_mqtt_server_fn(struct aws_allocator *allocator, void *ctx) {
...
+    aws_thread_current_sleep(ONE_MILLISECOND);
     uint64_t timestamp = 0;
     ASSERT_SUCCESS(aws_sys_clock_get_ticks(&timestamp));
 
     snprintf(
         state_test_data->endpoint.address,
         sizeof(state_test_data->endpoint.address),
         LOCAL_SOCK_TEST_PATTERN,
-        (long long unsigned)timestamp);
+        (long long unsigned)timestamp); // STEVE TODO maybe use a UUID here instead or add a UUID to the timestamp
```
TwistedTwigleg, line 256: Nitpick: Remove TODO?

TwistedTwigleg (commented):
Looks like it is failing due to a compiler issue missing "AWS_RESET_RECONNECT_BACKOFF_DELAY_SECONDS"? Not sure why though, as it looks like it is defined...

tests/v3/connection_state_test.c
sbSteveK: oops!

tests/v3/connection_state_test.c
sbSteveK: I forgot the ';' on line 29 >,< Fixed.

TwistedTwigleg (approved):

tests/v3/connection_state_test.c
graebm: I realize we have LOTS of tests, across lots of C repos with this pattern (use timestamp as socket name)

Your TODO is right, that we'd be better off using UUID, instead of a timestamp.
It's possible to run tests in parallel, where this sleep trick might not work, it would still be possible for multiple tests to be hitting this same line of code within the same millisecond.

Anyway, if this fixes your flaky test for now ... whatever. But if we keep encountering this kind of thing in the future, the UUID approach would be better

PR #279, opened 2023-05-01
Adjust the canary not to crash if it cannot poll the metric alarms

TwistedTwigleg:
*Description of changes:*

Minor canary script fix: If the canary cannot, for whatever reason, pull the state of the Canary status alarms, instead of crashing it will just skip and try again later. This is to prevent the canary from going down when there is too much network congestion or the service is temporarily down for a little bit.

_______

### DIFF
--- a/codebuild/CanaryWrapper_Classes.py
+++ b/codebuild/CanaryWrapper_Classes.py
@@ -67,14 +67,14 @@ class DataSnapshot_Dashboard_Widget():
             self.metric_list.append(new_metric_name)
         except Exception as e:
             print ("[DataSnapshot_Dashboard] ERROR - could not add metric to dashboard widget due to exception!")
-            print ("[DataSnapshot_Dashboard] Exception: " + str(e))
+            print (f"[DataSnapshot_Dashboard] Exception: {repr(e)}")
 
     def remove_metric_from_widget(self, existing_metric_name):
         try:
             self.metric_list.remove(existing_metric_name)
         except Exception as e:
             print ("[DataSnapshot_Dashboard] ERROR - could not remove metric from dashboard widget due to exception!")
-            print ("[DataSnapshot_Dashboard] Exception: " + str(e))
+            print (f"[DataSnapshot_Dashboard] Exception: {repr(e)}")
 
     def get_widget_dictionary(self):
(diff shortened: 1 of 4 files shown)


codebuild/CanaryWrapper_Classes.py
```diff
@@ -830,8 +830,7 @@ def monitor_loop_function(self, psutil_process : psutil.Process, time_passed=30)
         except Exception as e:
             self.print_message("[SnaptshotMonitor] ERROR - exception occurred checking metric alarms!")
             self.print_message("[SnaptshotMonitor] (Likely session credentials expired)")
-            self.had_internal_error = True
-            self.internal_error_reason = "Exception occurred checking metric alarms! Likely session credentials expired"
+            self.print_message("[SnaptshotMonitor] Not going to crash - just going to try again later")
```
bretambrose, line 833: A little confused about the session credentials issue.  If providers are being used properly, expiration should not be a concern.

codebuild/CanaryWrapper_Classes.py
```diff
@@ -840,8 +838,9 @@ def monitor_loop_function(self, psutil_process : psutil.Process, time_passed=30)
                     self.data_snapshot.post_metrics(psutil_process)
                 except Exception as e:
                     self.print_message("[SnaptshotMonitor] ERROR - exception occurred posting metrics!")
```
sbSteveK, line 840: Trivial: Is there any additional information we can extract from the exception that we can share in the logging?

sbSteveK (approved):

codebuild/CanaryWrapper_Classes.py
bretambrose: Well just fyi, they are session-based credentials being used on EC2 (presumably imds-based).  But the provider will refresh because expiration is tracked as well.

codebuild/CanaryWrapper_Classes.py
TwistedTwigleg: We could print the exception, since the logs will only ever be visible in Codebuild, there shouldn't be any credential leaking risks since only the team can see the Codebuild output.

TwistedTwigleg:

PR #283, opened 2023-05-18
Use UUID for socket endpoints

TwistedTwigleg:
*Description of changes:*

Adjust tests to use UUID instead of timestamps for socket endpoints, preventing any overlap if two tests run in parallel or at the exact same timestamp clock sample.

_________

### DIFF
--- a/tests/v3/connection_state_test.c
+++ b/tests/v3/connection_state_test.c
@@ -17,7 +17,6 @@
 #include <aws/testing/aws_test_harness.h>
 
 #include <aws/common/math.h>
-#include <aws/common/uuid.h>
 
 static const int TEST_LOG_SUBJECT = 60000;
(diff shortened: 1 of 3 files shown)


bretambrose (commented):
I feel like we should have kept everything the same except change the pattern format string from llu to PRInSTR

TwistedTwigleg:
If we keep it the same though, wouldn't that still have the issue where tests running in parallel or during the same milliseconds would have the same socket endpoint?

bretambrose:
> If we keep it the same though, wouldn't that still have the issue where tests running in parallel or during the same milliseconds would have the same socket endpoint?

They'd use the uuid

bretambrose (approved):

TwistedTwigleg:

PR #292, opened 2023-05-24
Some misc. cleanup

bretambrose:
* Remove the unused local subscribe API and logic
* Rename connection to adapter in places to make clear the distinction between the implementation (adapter) and the polymorphic base (connection)
* Remove support for building without websockets.  This hasn't worked in over two years and is not something we want to support going forward.

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -11,6 +11,8 @@
 #include <aws/mqtt/private/topic_tree.h>
 
 #include <aws/http/proxy.h>
+#include <aws/http/request_response.h>
+#include <aws/http/websocket.h>
 
 #include <aws/io/channel_bootstrap.h>
 #include <aws/io/event_loop.h>
@@ -22,11 +24,6 @@
 #include <aws/common/task_scheduler.h>
 
 #include <inttypes.h>
-
-#ifdef AWS_MQTT_WITH_WEBSOCKETS
-#    include <aws/http/request_response.h>
-#    include <aws/http/websocket.h>
-#endif
 
 #ifdef _MSC_VER
 #    pragma warning(disable : 4204)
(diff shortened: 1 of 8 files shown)


source/client.c
```diff
@@ -2737,58 +2489,54 @@ static enum aws_mqtt_client_request_state s_unsubscribe_send(
...
-         * "wins", does its logic, and then breaks the connection between the two.
-         */
-        task_arg->timeout_wrapper.timeout_task_arg = timeout_task_arg;
-        timeout_task_arg->task_arg_wrapper = &task_arg->timeout_wrapper;
+    if (aws_channel_slot_send_message(task_arg->connection->slot, message, AWS_CHANNEL_DIR_WRITE)) {
+        goto handle_error;
     }
 
+    /* TODO: timing should start from the message written into the socket, which is aws_io_message->on_completion
+     * invoked, but there are bugs in the websocket handler (and maybe also the h1 handler?) where we don't properly
```
sbSteveK, line 2518: I don't immediately see a ticket related to this. Should we open one if this is a bug?

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -100,12 +100,12 @@ static void s_set_interruption_handlers_task_fn(struct aws_task *task, void *arg
...
     }
 
-    struct aws_mqtt_client_connection_5_impl *connection = set_task->connection->impl;
+    struct aws_mqtt_client_connection_5_impl *adapter = set_task->connection->impl;
 
-    connection->on_interrupted = set_task->on_interrupted;
-    connection->on_interrupted_user_data = set_task->on_interrupted_user_data;
-    connection->on_resumed = set_task->on_resumed;
-    connection->on_resumed_user_data = set_task->on_resumed_user_data;
+    adapter->on_interrupted = set_task->on_interrupted;
```
sbSteveK, line 105: I really like this renaming

sbSteveK (approved):

source/client.c
bretambrose: It's just an indentation change and I have no desire to monkey with it.  But we did fix websocket completion callbacks during the mqtt5 implementation

PR #293, opened 2023-05-24
311 adapter lifecycle foundations 

bretambrose:
Misc. functionality needed to support connect/disconnect and lifecycle events in the 311_to_5 adapter

* Adds support for internal observers to reset the mqtt5 client's connection when it is established or being established.
* Reworks how ref-counting and state changes are protected and adds a simple helper function to safely make callbacks based on synchronized state
* Splits ref-counting into external and internal.  Internal is all ref-counts made inside the implementation in order to extend the lifetime of the adapter long enough to safely handle asynchronous operations.  External is ref counts made by users.  The adapter now follows a contract where as soon as the last external ref count is released, no more callbacks to user code will be made under any circumstances.  This greatly simplifies shutdown from a user perspective (it definitely complicates things internally).
* Move the adapter's locking approach to a read-write lock rather than a mutex.


source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -22,13 +51,61 @@ struct aws_mqtt_client_connection_5_impl {
...
+     * It also provides a nice solution for the fact that we cannot safely upgrade a read lock to a write lock.
+     */
+    bool in_synchronous_callback;
+
+    /*
+     * What state the mqtt311 interface wants to be in.  Always either AWS_MCS_CONNECTED or AWS_MCS_STOPPED.
+     * Tracked separately from the underlying mqtt5 client state in order to provide a coherent view that is
+     * narrowed solely to the adapter interface.
+     */
+    enum aws_mqtt_adapter_state adapter_state;
```
sbSteveK, line 68: should this enum be `aws_mqtt5_client_state`?

sbSteveK: nvm, this is used below intentionally as `aws_mqtt_adapter_state`. Maybe the description needs to be changed to reflect the enum values available for it( `AWS_MQTT_AS_STAY_CONNECTED` instead of `AWS_MCS_CONNECTED` )? Or clarified that the mqtt5 client is always in either the listed `AWS_MCS_CONNECTED` or `AWS_MCS_STOPPED` desired states? Trivial but just for clarification.

bretambrose: Updated the out-of-date comment

PR #296, opened 2023-06-01
Adapter lifecycle take2

bretambrose:
Mqtt3 to 5 adapter lifecycle implementation

Maps the MQTT5 lifecycle event set as best it can to the MQTT311 implementation's event set.  Given that the two interfaces can undermine one another, this is a best effort implementation; there are multiple situations where we have to make a judgement call.


include/aws/mqtt/private/v5/mqtt5_client_impl.h
```diff
@@ -673,6 +673,19 @@ AWS_MQTT_API const char *aws_mqtt5_client_state_to_c_string(enum aws_mqtt5_clien
  */
 AWS_MQTT_API bool aws_mqtt5_client_reset_connection(struct aws_mqtt5_client *client);
 
+/**
+ * Event-loop-internal API used to switch the client's desired state.  Used by both start() and stop() cross-thread
+ * tasks as well as by the 3-to-5 adapter to make changes synchronously (when in the event loop).
+ *
+ * @param client mqtt5 client to update desired state for
+ * @param desired_state new desired state
+ * @param disconnect_op optional description of a DISCONNECT packet to send as part of a stop commandf
```
sbSteveK, line 682: trivial: "command"

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -260,8 +269,451 @@ static int s_aws_mqtt5_adapter_perform_safe_callback(
...
+    connect_task->port = connection_options->port;
+    connect_task->socket_options = *connection_options->socket_options;
+    if (connection_options->tls_options) {
+        aws_tls_connection_options_copy(&connect_task->tls_options, connection_options->tls_options);
+        connect_task->tls_options_ptr = &connect_task->tls_options;
+    }
+
+    aws_byte_buf_init_copy_from_cursor(&connect_task->client_id, allocator, connection_options->client_id);
+
+    connect_task->keep_alive_time_secs = connection_options->keep_alive_time_secs;
```
sbSteveK, line 382: Tangent: If we aren't already elsewhere, should we force/change/check that the ping_timeout_ms is less than the keep_alive_time_secs?

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -260,8 +269,451 @@ static int s_aws_mqtt5_adapter_perform_safe_callback(
...
+
+static int s_aws_mqtt_client_connection_5_connect(
+    void *impl,
+    const struct aws_mqtt_connection_options *connection_options) {
+
+    struct aws_mqtt_client_connection_5_impl *adapter = impl;
+
+    struct aws_mqtt_adapter_connect_task *task =
+        s_aws_mqtt_adapter_connect_task_new(adapter->allocator, adapter, connection_options);
+    if (task == NULL) {
```
sbSteveK, line 400: Tangent: Can these fail in a way that returns NULL? Just wondering if this is a check and error reporting based on back when aws_mem_calloc could fail.

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -260,8 +269,451 @@ static int s_aws_mqtt5_adapter_perform_safe_callback(
...
+    /*
+     * Never invoke a callback after termination
+     */
+    if (adapter->synced_data.terminated) {
+        return AWS_OP_SUCCESS;
+    }
+
+    switch (event->event_type) {
+
+        case AWS_MQTT5_CLET_CONNECTION_SUCCESS:
```
sbSteveK, line 424: Just noting the obvious that we'll need to add the new mqtt3 callbacks into this section once they're implemented elsewhere -_-;;

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -260,8 +269,451 @@ static int s_aws_mqtt5_adapter_perform_safe_callback(
...
+            if (adapter->on_closed) {
+                (*adapter->on_closed)(&adapter->base, NULL, adapter->on_closed_user_data);
+            }
+
+            /*
+             * Judgement call: If the mqtt5 client is stopped behind our back, it seems better to transition to the
+             * disconnected state (which only requires a connect() to restart) then stay in the STAY_CONNECTED state
+             * which currently requires a disconnect() and then a connect() to restore connectivity.
+             *
+             * ToDo: what if we disabled mqtt5 client start/stop somehow while the adapter is attached, preventing
```
sbSteveK, line 512: If we take this path, both start() and stop() for mqtt5 return an int that's currently just whether or not s_aws_mqtt5_client_change_desired_state() was successfully scheduled. Could we add a check in that function or start() and stop() itself and return/log a new error that clearly states that start() and/or stop() aren't available when an adapter is attached and not allow the user to call them?

source/mqtt3_to_mqtt5_adapter.c
sbSteveK: Ah, we do this validation in the aws_mqtt5_client_options_validate() but I think this implementation currently bypasses that validation and directly sets the stored config. Should we process these options through validation or just allow things to potentially fail on mqtt5 attempting to connect with bad settings?

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -260,8 +269,451 @@ static int s_aws_mqtt5_adapter_perform_safe_callback(
...
+    }
+
+    adapter->adapter_state = AWS_MQTT_AS_FIRST_CONNECT;
+
+    /* Update mqtt5 config */
+    s_aws_mqtt3_to_mqtt5_adapter_update_config_on_connect(adapter, connect_task);
+
+    aws_mqtt5_client_reset_connection(adapter->client);
+
+    aws_mqtt5_client_change_desired_state(adapter->client, AWS_MCS_CONNECTED, NULL);
```
sbSteveK, line 692: Mentioned above, I might be missing something but here we are changing the stored config options, I think without validating that they're good, and then directly changing the state to initiate a connect.

tests/v5/mqtt5_client_tests.c
```diff
@@ -97,16 +97,9 @@ void s_publish_received_callback(const struct aws_mqtt5_packet_publish_view *pub
     (void)user_data;
 }
 
-AWS_STATIC_STRING_FROM_LITERAL(s_client_id, "HelloWorld");
+AWS_STRING_FROM_LITERAL(g_default_client_id, "HelloWorld");
```
sbSteveK, line 100: This is set here but not in mqtt3_to_mqtt5_adapter_tests.c. g_default_client_id is extern const and I'm not sure if that means setting this here will insure it's set when the adapter tests run.

source/mqtt3_to_mqtt5_adapter.c
sbSteveK: After recent discussion, this doesn't make much sense since the mqtt5 client could be being used as a straight up mqtt5 client.

sbSteveK (approved):

source/mqtt3_to_mqtt5_adapter.c
bretambrose: Sure, will add

source/mqtt3_to_mqtt5_adapter.c
bretambrose: Maybe, but the task construction is complex enough that I don't think we should be transitively assuming so.

source/mqtt3_to_mqtt5_adapter.c
bretambrose: Added keep alive validation for now, can consider other fields as well, but doesn't seem pressing.

tests/v5/mqtt5_client_tests.c
bretambrose: extern is a declaration not a definition

source/mqtt3_to_mqtt5_adapter.c
```diff
@@ -260,8 +269,451 @@ static int s_aws_mqtt5_adapter_perform_safe_callback(
...
+                AWS_ERROR_MQTT_ALREADY_CONNECTED,
+                0,
+                false,
+                connect_task->on_connection_complete_user_data);
+        }
+        return AWS_OP_SUCCESS;
+    }
+
+    if (adapter->on_disconnect) {
+        (*adapter->on_disconnect)(&adapter->base, adapter->on_disconnect_user_data);
```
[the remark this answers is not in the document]
bretambrose: The disconnect callback is transient.  In the 311 interface it corresponds to a single call to disconnect()

source/mqtt3_to_mqtt5_adapter.c
bretambrose: Added some additional validation

PR #303, opened 2023-07-06
Adapter subscribe unsubscribe

bretambrose:
Adds implementations and basic tests for subscribe, subscribe_multi, publish, and unsubscribe to the 3-to-5 adapter.


include/aws/mqtt/private/v5/mqtt3_to_mqtt5_adapter_impl.h
```diff
@@ -78,6 +94,46 @@ struct aws_mqtt3_to_mqtt5_adapter_operation_publish {
...
+    /*
+     * holds a reference to the MQTT5 client subscribe operation until the operation completes or our adapter
+     * goes away.
+     *
+     * In the case where we're going away, we zero out the MQTT5 operation callbacks to prevent crash-triggering
+     * notifications.
+     */
+    struct aws_mqtt5_operation_subscribe *subscribe_op;
+
+    /* aws_array_list<struct aws_mqtt_subscription_set_subscription_record *> */
```
sbSteveK, line 109: delete?

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1819,12 +2393,26 @@ void aws_mqtt3_to_mqtt5_adapter_operation_table_remove_operation(
...
+    }
+}
+
+struct aws_mqtt3_to_mqtt5_adapter_operation_base *aws_mqtt3_to_mqtt5_adapter_operation_release(
+    struct aws_mqtt3_to_mqtt5_adapter_operation_base *operation) {
+    if (operation != NULL) {
+        aws_ref_count_release(&operation->ref_count);
     }
+
+    return NULL;
```
sbSteveK, line 2406: Trivial: Is this required?

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1819,12 +2393,26 @@ void aws_mqtt3_to_mqtt5_adapter_operation_table_remove_operation(
...
+struct aws_mqtt3_to_mqtt5_adapter_operation_base *aws_mqtt3_to_mqtt5_adapter_operation_acquire(
+    struct aws_mqtt3_to_mqtt5_adapter_operation_base *operation) {
+    if (operation != NULL) {
+        aws_ref_count_acquire(&operation->ref_count);
+    }
+
+    return operation;
 }
 
 void aws_mqtt3_to_mqtt5_adapter_operation_reference_adapter(
```
sbSteveK, line 2418: Trivial: I think we use acquire and release for referencing and dereferencing. Should we stick to that naming convention for readability? I felt like I knew what was going on in this function from the context but had to check to be sure because it was doing the same thing but named different from the immediately following line up at line 1667 to 1670.

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1651,7 +1681,549 @@ static uint16_t s_aws_mqtt_client_connection_5_publish(
...
+    if (subscribe_op->on_multi_suback != NULL) {
+        AWS_VARIABLE_LENGTH_ARRAY(
+            struct aws_mqtt_topic_subscription, multi_sub_subscription_buf, suback->reason_code_count);
+        AWS_VARIABLE_LENGTH_ARRAY(
+            struct aws_mqtt_topic_subscription *, multi_sub_subscription_ptr_buf, suback->reason_code_count);
+        struct aws_mqtt_topic_subscription *subscription_ptr =
+            (struct aws_mqtt_topic_subscription *)multi_sub_subscription_buf;
+
+        struct aws_array_list multi_sub_list;
+        aws_array_list_init_static(
```
sbSteveK, line 1783: Yikes, this was a rabbit hole. Will have to take note to use this pattern in the future if I need a temp array list >,<

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1651,7 +1681,549 @@ static uint16_t s_aws_mqtt_client_connection_5_publish(
...
+
+    aws_mqtt3_to_mqtt5_adapter_operation_table_remove_operation(
+        &subscribe_op->base.adapter->operational_state, subscribe_op->base.id);
+}
+
+static int s_validate_adapter_subscribe_options(const struct aws_mqtt3_to_mqtt5_adapter_subscribe_options *options) {
+    if (options->subscription_count == 0) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
```
sbSteveK, line 1826: `s_aws_mqtt_client_connection_5_unsubscribe` runs the topics through `aws_mqtt_is_valid_topic_filter` but I don't think subscribe or subscribe multiple does. Maybe we can add it to those or here?

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1651,7 +1681,549 @@ static uint16_t s_aws_mqtt_client_connection_5_publish(
...
+    subscribe_op->base.allocator = allocator;
+    aws_ref_count_init(&subscribe_op->base.ref_count, subscribe_op, s_adapter_subscribe_operation_destroy);
+    subscribe_op->base.impl = subscribe_op;
+    subscribe_op->base.type = AWS_MQTT3TO5_AOT_SUBSCRIBE;
+    subscribe_op->base.adapter = options->adapter;
+    subscribe_op->base.holding_adapter_ref = false;
+
+    /* make persistent adapter sub array */
+    aws_array_list_init_dynamic(
```
sbSteveK, line 1857: `aws_array_list_init_dynamic` looks like it can fail? Or at least that possibility is handled in a lot of other areas. Should we check its result and goto error here?

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1651,7 +1681,549 @@ static uint16_t s_aws_mqtt_client_connection_5_publish(
...
+        aws_array_list_get_at(&operation->subscriptions, &record, i);
+
+        aws_mqtt_subscription_set_add_subscription(adapter->subscriptions, &record->subscription_view);
+    }
+
+    aws_mqtt5_client_submit_operation_internal(
+        adapter->client, &operation->subscribe_op->base, status != AWS_TASK_STATUS_RUN_READY);
+
+    /* release the transient adapter reference for the move to the event loop */
+    aws_mqtt3_to_mqtt5_adapter_operation_dereference_adapter(&operation->base);
```
sbSteveK, line 1940: Trivial: The more I'm seeing it, the stronger I'm feeling about renaming this to `aws_mqtt3_to_mqtt5_adapter_operation_adapter_release` to match what we're using just below with `aws_mqtt3_to_mqtt5_adapter_operation_release`. Haha, actually writing it out though, the repeat use of the word adapter and its extreme similarity to that other release call does also feel weird. I dunno.

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1651,7 +1681,549 @@ static uint16_t s_aws_mqtt_client_connection_5_publish(
...
+    }
+
+    if (aws_mqtt3_to_mqtt5_adapter_operation_table_add_operation(&adapter->operational_state, &operation->base)) {
+        goto error;
+    }
+
+    uint16_t synthetic_id = operation->base.id;
+
+    /* add a transient reference to the adapter until this operation is passed into the event loop */
+    aws_mqtt3_to_mqtt5_adapter_operation_reference_adapter(&operation->base);
```
sbSteveK, line 2040: Trivial: Apologize for being so hung up on this but it's like a splinter in my brain now. `aws_mqtt3_to_mqtt5_adapter_operation_reference_adapter` and `aws_mqtt3_to_mqtt5_adapter_operation_acquire` are always called one after the other and both work off the same base. Would it make any sense at all to handle both releases in a singular function that handles the transient references? Just a thought since this is a repeating pattern in 4 places.

include/aws/mqtt/private/v5/mqtt3_to_mqtt5_adapter_impl.h
bretambrose: It's communicating what the container type is since we don't have generics in C

source/v5/mqtt3_to_mqtt5_adapter.c
bretambrose: It's the ref counting contract that I like since it allows you to minimally swap or drop temporary references without worrying about NULL.

In particular, assigning the result of acquire/release is always correct and safe.  If your release is in your destructor, then it doesn't add much, but if your release is mid-lifetime of an object, it simplifies the correct logic (you need to assign to NULL as well).

source/v5/mqtt3_to_mqtt5_adapter.c
bretambrose: Updated to use a cross_thread_refs acquire and release function pair

source/v5/mqtt3_to_mqtt5_adapter.c
bretambrose: Memory allocation only

source/v5/mqtt3_to_mqtt5_adapter.c
bretambrose: That would result in a validation failure which would fail the overall request.  I'll add a test that does a invalid subscribe though.

bretambrose: Actually, thinking about it more, that should be a synchronous failure on all of unsubscribe/subscribe/subscribe-multi because that's the way the 311 impl handles it.

sbSteveK (approved):

source/v5/mqtt3_to_mqtt5_adapter.c
bretambrose: NM it already is a synchronous failure when the MQTT5 operation creation fails due to validation.

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1506,6 +1506,26 @@ static void s_aws_mqtt_client_connection_5_release(void *impl) {
     aws_ref_count_release(&adapter->external_refs);
 }
 
+static void s_aws_mqtt3_to_mqtt5_adapter_operation_acquire_cross_thread_refs(
```
sbSteveK, line 1509: 🙌

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1663,11 +1678,7 @@ static uint16_t s_aws_mqtt_client_connection_5_publish(
 
     uint16_t synthetic_id = operation->base.id;
 
-    /* transient ref on adapter while moving to the event loop thread */
-    aws_mqtt3_to_mqtt5_adapter_operation_reference_adapter(&operation->base);
-
-    /* transient ref on operation while moving to the event loop thread */
-    aws_mqtt3_to_mqtt5_adapter_operation_acquire(&operation->base);
+    s_aws_mqtt3_to_mqtt5_adapter_operation_acquire_cross_thread_refs(&operation->base);
```
sbSteveK, line 1681: Absolute nitpicking but should we leave the comment about why this is happening? It seemed useful.

sbSteveK (approved):

PR #305, opened 2023-07-14
Adapter resub stats

bretambrose:
* get stats implementation & test
* resubscribe implementation & tests


include/aws/mqtt/private/v5/mqtt3_to_mqtt5_adapter_impl.h
```diff
@@ -29,6 +29,11 @@ struct aws_mqtt3_to_mqtt5_adapter_publish_options {
     void *on_complete_userdata;
 };
 
+/*
+ * A subscribe with no subscriptions represents a re-subscribe of all internally tracked topics.  While this
```
sbSteveK, line 33: Us being the only ones to ever directly run across this, this seems fine. Especially since it's well documented what's going on for future poor souls digging around!

source/v5/mqtt3_to_mqtt5_adapter.c
```diff
@@ -1907,10 +1896,39 @@ struct aws_mqtt3_to_mqtt5_adapter_operation_subscribe *aws_mqtt3_to_mqtt5_adapte
...
+    subscribe_op->base.impl = subscribe_op;
+    subscribe_op->base.type = AWS_MQTT3TO5_AOT_SUBSCRIBE;
+    subscribe_op->base.adapter = options->adapter;
+    subscribe_op->base.holding_adapter_ref = false;
+
+    /*
+     * If we're a regular subscribe, build the mqtt5 operation now.  Otherwise, we have to wait until
+     * we're on the event loop thread and it's safe to query the subscription set.
+     */
+    if (options->subscription_count > 0) {
```
sbSteveK, line 1927: If it's possible to call `aws_mqtt_client_connection_subscribe_multiple` with an empty aws_array_list, the subscription_count would be zero, the subscribe_op->subscribe_op would remain null, and the subscribe multiple might trigger a resubscribe.

sbSteveK (approved):

source/v5/mqtt3_to_mqtt5_adapter.c
bretambrose: True, will add a synchronous failure check for that case.

PR #311, opened 2023-07-24
Mqtt311 decoder

bretambrose:
* Fix a framing bug that led to dead/broken connections
* Refactor the framing logic used by the mqtt311 implementation to use an encapsulated, testable subsystem (aws_mqtt311_decoder).

While we could have fixed the original bug without doing such a heavyweight refactor, testing complex, brittle logic embedded in the channel handler's vtable implementation is a lousy approach.  Rather than graft more hacks and more complexity into the client impl, we add a new subsystem with a testable API.


source/client_channel_handler.c
```diff
@@ -219,9 +249,12 @@ static int s_packet_handler_connack(
     return AWS_OP_SUCCESS;
 }
 
-static int s_packet_handler_publish(
-    struct aws_mqtt_client_connection_311_impl *connection,
-    struct aws_byte_cursor message_cursor) {
+static int s_packet_handler_publish(struct aws_byte_cursor message_cursor, void *user_data) {
+    struct aws_mqtt_client_connection_311_impl *connection = user_data;
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: PUBLISH received", (void *)connection);
```
sbSteveK, line 254: Trivial: Should we log reception of PUBLISH at a debug level?

include/aws/mqtt/private/fixed_header.h
```diff
@@ -59,4 +59,6 @@ AWS_MQTT_API int aws_mqtt_fixed_header_encode(struct aws_byte_buf *buf, const st
  */
 AWS_MQTT_API int aws_mqtt_fixed_header_decode(struct aws_byte_cursor *cur, struct aws_mqtt_fixed_header *header);
 
+AWS_MQTT_API int aws_mqtt311_decode_remaining_length(struct aws_byte_cursor *cur, size_t *remaining_length_out);
```
sbSteveK, line 62: Trivial: MQTT5's variable byte integer ( https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901011 ) is encoded the same way as MQTT3's variable length encoding ( http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023 ). Just wondering if we should unify the decoding of VLIs. Not necessary but feels funny that we have two things doing the same thing differently. 

MQTT3
https://github.com/awslabs/aws-c-mqtt/blob/7aaae0ce8c644faf12fcb6544f8f105787596101/source/fixed_header.c#L32C27-L32C27

MQTT5
https://github.com/awslabs/aws-c-mqtt/blob/7aaae0ce8c644faf12fcb6544f8f105787596101/source/v5/mqtt5_decoder.c#L75

bretambrose: In the future maybe, but the contracts are different and I didnt want to unify them as part of this.

source/client_channel_handler.c
bretambrose: I was going to add the log level refactor into this but later reconsidered.  But yes, that one should be debug.

source/mqtt311_decoder.c
```diff
@@ -0,0 +1,206 @@
...
+     * If we successfully decoded a variable-length integer, we now know exactly how many bytes we need to receive in
+     * order to have the full packet.
+     */
+    if (vli_result == AWS_OP_SUCCESS) {
+        decoder->total_packet_length = remaining_length + decoder->packet_buffer.len;
+        AWS_FATAL_ASSERT(decoder->total_packet_length > 0);
+        decoder->state = AWS_MDST_READ_BODY;
+    }
+
+    return AWS_MDD_CONTINUE;
```
graebm, line 137: trivial/style: I'd find the logic easier to follow if it were a bunch of if/elses. vs if-unexpected-error, if-success, implicitly-else this is an errror we were expecting
```suggestion
    if (aws_mqtt311_decode_remaining_length(&vli_cursor, &remaining_length) == AWS_OP_ERR) {
        if (aws_last_error() == AWS_ERROR_SHORT_BUFFER) {
            /* not enough data yet */
            return AWS_MDD_CONTINUE;
        } else {
            /* anything else is a terminal error */
            return AWS_MDD_PROTOCOL_ERROR;
        }
    } else {
        /* We successfully decoded a variable-length integer, we now know
         * exactly how many bytes we need to receive in order to have the full packet. */
        decoder->total_packet_length = remaining_length + decoder->packet_buffer.len;
        AWS_FATAL_ASSERT(decoder->total_packet_length > 0);
        decoder->state = AWS_MDST_READ_BODY;
    }
```

source/mqtt311_decoder.c
```diff
@@ -0,0 +1,206 @@
...
+    struct aws_mqtt311_decoder *decoder,
+    struct aws_byte_cursor *data) {
+    AWS_FATAL_ASSERT(decoder->total_packet_length > 0);
+
+    size_t buffer_length = decoder->packet_buffer.len;
+    size_t amount_to_read = aws_min_size(decoder->total_packet_length - buffer_length, data->len);
+
+    struct aws_byte_cursor copy_cursor = aws_byte_cursor_advance(data, amount_to_read);
+    if (aws_byte_buf_append_dynamic(&decoder->packet_buffer, &copy_cursor)) {
+        return AWS_MDD_PROTOCOL_ERROR;
```
graebm, line 150: trivial: you're checking this append_dynamic, but not others

I'd remove the error-checking, failed allocations are fatal

graebm (commented):
https://www.youtube.com/watch?v=NoDsJAiseGg

bretambrose:
> https://www.youtube.com/watch?v=NoDsJAiseGg

Even better: https://www.youtube.com/watch?v=mjfT2joaJxs

source/mqtt311_decoder.c
```diff
@@ -0,0 +1,206 @@
...
+     * states if the current MQTT packet was received in a fragmented manner.
+     */
+    struct aws_byte_cursor temp_cursor = *data;
+    struct aws_mqtt_fixed_header packet_header;
+    AWS_ZERO_STRUCT(packet_header);
+    if (!aws_mqtt_fixed_header_decode(&temp_cursor, &packet_header) &&
+        temp_cursor.len >= packet_header.remaining_length) {
+
+        /* figure out the cursor that spans the full packet */
+        size_t fixed_header_length = temp_cursor.ptr - data->ptr;
```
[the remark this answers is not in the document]
bretambrose: I'm don't think that's equivalent, data is unchanged at this point

source/mqtt311_decoder.c
```diff
@@ -0,0 +1,206 @@
...
+     * full decoder state machine.
+     *
+     * A corollary of this is that the decoder is only ever in the AWS_MDST_READ_REMAINING_LENGTH or AWS_MDST_READ_BODY
+     * states if the current MQTT packet was received in a fragmented manner.
+     */
+    struct aws_byte_cursor temp_cursor = *data;
+    struct aws_mqtt_fixed_header packet_header;
+    AWS_ZERO_STRUCT(packet_header);
+    if (!aws_mqtt_fixed_header_decode(&temp_cursor, &packet_header) &&
+        temp_cursor.len >= packet_header.remaining_length) {
```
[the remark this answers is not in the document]
bretambrose: Every packet decode invokes it, so it's going to get caught in the buffered path, just not on the cursor shortcut (which fails and forces the buffered path).  I'm not sure the extra complexity is warranted but start with adding a failure test for packets with bad flags.

source/client_channel_handler.c
```diff
@@ -94,11 +89,46 @@ static void s_on_time_to_ping(struct aws_channel_task *channel_task, void *arg,
...
+                "id=%p: First message received from the server was not a CONNACK. Terminating connection.",
+                (void *)connection);
+            aws_channel_shutdown(connection->slot->channel, AWS_ERROR_MQTT_PROTOCOL_ERROR);
+            return aws_raise_error(AWS_ERROR_MQTT_PROTOCOL_ERROR);
+        }
+        mqtt_connection_unlock_synced_data(connection);
+    } /* END CRITICAL SECTION */
 
-    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: CONNACK received", (void *)connection);
+    if (AWS_UNLIKELY(packet_type > AWS_MQTT_PACKET_DISCONNECT || packet_type < AWS_MQTT_PACKET_CONNECT)) {
```
graebm, line 112: shouldn't this check be in the decoder, since the callback we invoke is based on packet_type? so an unexpected packet_type would be an invalid array access?

source/mqtt311_decoder.c
bretambrose: switched, but dropped the outer else

source/mqtt311_decoder.c
```diff
@@ -0,0 +1,206 @@
...
+                decode_directive = AWS_MDD_PROTOCOL_ERROR;
+                break;
+        }
+
+        /*
+         * Protocol error is a terminal failure state until aws_mqtt311_decoder_reset_for_new_connection() is called.
+         */
+        if (decode_directive == AWS_MDD_PROTOCOL_ERROR) {
+            decoder->state = AWS_MDST_PROTOCOL_ERROR;
+            return aws_raise_error(AWS_ERROR_MQTT_PROTOCOL_ERROR);
```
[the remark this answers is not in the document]
bretambrose: This was more me disliking short buffer getting set as an error under completely normal circumstances.  I'll just not override.

source/client_channel_handler.c
```diff
@@ -94,11 +89,46 @@ static void s_on_time_to_ping(struct aws_channel_task *channel_task, void *arg,
...
+        mqtt_connection_lock_synced_data(connection);
+        /* [MQTT-3.2.0-1] The first packet sent from the Server to the Client MUST be a CONNACK Packet */
+        if (connection->synced_data.state == AWS_MQTT_CLIENT_STATE_CONNECTING &&
+            packet_type != AWS_MQTT_PACKET_CONNACK) {
+            mqtt_connection_unlock_synced_data(connection);
+            AWS_LOGF_ERROR(
+                AWS_LS_MQTT_CLIENT,
+                "id=%p: First message received from the server was not a CONNACK. Terminating connection.",
+                (void *)connection);
+            aws_channel_shutdown(connection->slot->channel, AWS_ERROR_MQTT_PROTOCOL_ERROR);
```
graebm, line 106: this error-path calls channel_shutdown() but other error-paths don't

probably simpler to just raise an error here, and let channel_shutdown() get called after aws_mqtt311_decoder_on_bytes_received() fails. Vs calling channel_shutdown() is some places you raise an error, and not others, and then call channel_shutdown() should get called in the end anyway

source/client_channel_handler.c
```diff
@@ -94,11 +89,46 @@ static void s_on_time_to_ping(struct aws_channel_task *channel_task, void *arg,
...
-static int s_packet_handler_connack(
+
+static int s_validate_received_packet_type(
     struct aws_mqtt_client_connection_311_impl *connection,
-    struct aws_byte_cursor message_cursor) {
+    enum aws_mqtt_packet_type packet_type) {
+    { /* BEGIN CRITICAL SECTION */
+        mqtt_connection_lock_synced_data(connection);
+        /* [MQTT-3.2.0-1] The first packet sent from the Server to the Client MUST be a CONNACK Packet */
+        if (connection->synced_data.state == AWS_MQTT_CLIENT_STATE_CONNECTING &&
```
graebm, line 99: debatable: Maybe just have the decoder check that the first packet it ever receives is a CONNACK packet? and then we don't need this validation function that every packet-handler needs to call...

I just found it weird that it grabs the lock, checks the value of `state` but doesn't change it, then the calling s_packet_handler_connack() function grabs the lock again and changes the value of `state`

I realize the decoder can't validate everything related to the context in which a packet was received, but if it can validate basic stuff ... that seems worth doing, vs spreading the complexity out here

source/client_channel_handler.c
```diff
@@ -531,103 +595,15 @@ static int s_process_read_message(
...
-                AWS_LOGF_TRACE(
-                    AWS_LS_MQTT_CLIENT, "id=%p: message is incomplete, waiting on another read.", (void *)connection);
-                if (aws_byte_buf_init(
-                        &connection->thread_data.pending_packet,
-                        connection->allocator,
-                        fixed_header_size + packet_header.remaining_length)) {
-
-                    return AWS_OP_ERR;
-                }
+    int result = aws_mqtt311_decoder_on_bytes_received(&connection->thread_data.decoder, message_cursor);
```
graebm, line 598: might want to take more drastic action after a decoder failure, like putting the connection into a state where it refuses to process any more data. or have aws_mqtt311_decoder_on_bytes_received() first check that it's not already in an error state

aws_channel_shutdown() isn't immediate. Other channel handlers might have data queued up that they'll keep sending the the MQTT-handler. It would be pretty bad to report a bad packet, but then continue processing the next

source/client_channel_handler.c
```diff
@@ -531,103 +595,15 @@ static int s_process_read_message(
...
-        }
-
-        struct aws_byte_cursor packet_data =
-            aws_byte_cursor_advance(&message_cursor, fixed_header_size + packet_header.remaining_length);
-        AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: full mqtt packet read, dispatching.", (void *)connection);
-        s_process_mqtt_packet(connection, packet_header.packet_type, packet_data);
+    if (result == AWS_OP_SUCCESS) {
+        /* Do cleanup */
+        aws_channel_slot_increment_read_window(slot, message->message_data.len);
+        aws_mem_release(message->allocator, message);
```
graebm, line 603: trivial: reverse these 2 commands. better to put the message back in the pool before you say "ok you can send me more data now".

I'm pretty sure increment_read_window() won't recurse and give you more data (but it used to).

source/client_channel_handler.c
bretambrose: The safe packet handler wrapper checks for null.

source/client_channel_handler.c
bretambrose: Well once the decoder goes into a terminal state, it just always returns errors on incoming data

source/mqtt311_decoder.c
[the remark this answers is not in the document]
graebm: whatever, if my suggestion is confusing too, then it's not much better

source/mqtt311_decoder.c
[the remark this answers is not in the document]
graebm: I was just worried a bad flag would put us in an infinite loop, where this function keeps falling through to
```
    decoder->state = AWS_MDST_READ_REMAINING_LENGTH;
    return AWS_MDD_CONTINUE;
```

source/client_channel_handler.c
bretambrose: Didn't want to do anything other than a mechnical reworking of the packet handlers to support testing (don't need an actual connection).  Unsure if going further is a good idea from a safety standpoint.

source/client_channel_handler.c
graebm: currently, if you call aws_mqtt311_decoder_on_bytes_received() again after an error, it still processes the new data

source/mqtt311_decoder.c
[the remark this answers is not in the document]
xiazhvera: The `data` is unchanged, while `temp_cursor.len` has been updated since we proceed the header here. It seems that the difference of `data.len` and `temp_cursor.len` should be the same as the header length here.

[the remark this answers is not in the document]
bretambrose: NM I see it now.

source/client_channel_handler.c
bretambrose: Also, I don't really want to put protocol features in the decoder both from a separation of concerns standpoint and the fact that it would mess up all the tests if the decoder doesn't get a connack first.

source/client_channel_handler.c
bretambrose: Once it's in the protocol error state it just returns AWS_OP_ERR to subsequent calls.

source/client_channel_handler.c
graebm: sure, either way.
I worked around that in HTTP/2 like this:
```
    /* If true, do not expect the connection preface and immediately accept any frame type.
     * Only set this when testing the decoder itself */
    bool skip_connection_preface;
```

source/client_channel_handler.c
bretambrose: Gonna leave this in since it forces a protocol error and its the original stuff which I don't want to mess with any more than necessary.  But I'm going to do a second shutdown in the call to the decoder from the channel read message process function.

source/mqtt311_decoder.c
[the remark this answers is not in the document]
bretambrose: Added a test to check

sbSteveK (approved):

PR #313, opened 2023-08-02
Fix websocket transfer callback userdata & server name in tls_options

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/v5/mqtt5_to_mqtt3_adapter.c
+++ b/source/v5/mqtt5_to_mqtt3_adapter.c
@@ -1350,7 +1350,10 @@ static void s_aws_mqtt5_adapter_transform_websocket_handshake_fn(
         adapter->mqtt5_websocket_handshake_completion_user_data = complete_ctx;
 
         (*adapter->websocket_handshake_transformer)(
-            request, user_data, s_aws_mqtt5_adapter_websocket_handshake_completion_fn, adapter);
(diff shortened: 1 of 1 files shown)


source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -326,6 +340,13 @@ static struct aws_mqtt_adapter_connect_task *s_aws_mqtt_adapter_connect_task_new
     connect_task->clean_session = connection_options->clean_session;
 
     return connect_task;
+
+error:
+    aws_ref_count_release(&adapter->internal_refs);
```
bretambrose, line 345: Instead of doing this, let's move the acquire after the last control flow that can error out.  Having to release refs in error blocks is something to avoid if possible.

bretambrose (approved):
Fix n ship

source/v5/mqtt5_to_mqtt3_adapter.c
xiazhvera: The task was using the `adapter->allocator`. If we do not acquire the adapter beforehand, would there be chance that that the allocator get released with the adapter?

bretambrose: no

PR #314, opened 2023-08-08
Pushoff keep alive on ack received

xiazhvera:
*Issue #, if available:*

*Description of changes:*


bretambrose (changes requested):
The write timestamp can't be shared/singleton.

If I send 5 publishes at times 1, 2, 3, 4, 5.  And then an ack comes back for the first one, the push out needs to be based on the first one's write stamp (1).

xiazhvera:
> The write timestamp can't be shared/singleton.
> 
> If I send 5 publishes at times 1, 2, 3, 4, 5. And then an ack comes back for the first one, the push out needs to be based on the first one's write stamp (1).

I dont think it matters, we only need to push off based on the "last" write time since that's when our "last transmission" happened.

bretambrose:
> > The write timestamp can't be shared/singleton.
> > If I send 5 publishes at times 1, 2, 3, 4, 5. And then an ack comes back for the first one, the push out needs to be based on the first one's write stamp (1).
> 
> I dont think it matters, we only need to push off based on the "last" write time since that's when our "last transmission" happened.



> > The write timestamp can't be shared/singleton.
> > If I send 5 publishes at times 1, 2, 3, 4, 5. And then an ack comes back for the first one, the push out needs to be based on the first one's write stamp (1).
> 
> I dont think it matters, we only need to push off based on the "last" write time since that's when our "last transmission" happened.

It absolutely matters.  If the connect is dropped at time 1.5 then the ping cannot be pushed out farther than that.

tests/v3/connection_state_test.c
```diff
@@ -3366,8 +3366,8 @@ AWS_TEST_CASE_FIXTURE(
     &test_data)
 
 /**
- * Makes a CONNECT, with 1 second keep alive ping interval, send a publish roughly every second, and then ensure NO
- * pings were sent
+ * Makes a CONNECT, with 1 second keep alive ping interval, send a different operation every second, and then ensure NO
```
bretambrose, line 3369: If this comment is accurate then I feel like this test is likely overly sensitive to timing conditions and containerization.

tests/v3/connection_state_test.c
```diff
@@ -3387,25 +3387,120 @@ static int s_test_mqtt_connection_ping_no_fn(struct aws_allocator *allocator, vo
...
+    mqtt_connection_ping_no,
+    s_setup_mqtt_server_fn,
+    s_test_mqtt_connection_ping_no_fn,
+    s_clean_up_mqtt_server_fn,
+    &test_data)
+
+/**
+ * Makes a CONNECT, with 1 second keep alive ping interval, disable the server auto ack so that we never received ack
+ * back. Send a qos1 publish roughly every second for 4 seconds. As we never received the ACK, we should send a total of
+ * 4 ping
```
bretambrose, line 3457: This is alarming.  Given a keep aliveof 1 second and a ping timeout of 100ms we should be seeing multiple disconnections which are going to disrupt the logical simplicity of the test.

tests/v3/connection_state_test.c
```diff
@@ -3387,25 +3387,120 @@ static int s_test_mqtt_connection_ping_no_fn(struct aws_allocator *allocator, vo
...
         aws_thread_current_sleep(800000000);
     }
 
-    /* Ensure the server got 0 PING packets */
-    ASSERT_INT_EQUALS(0, mqtt_mock_server_get_ping_count(state_test_data->mock_server));
+    /*
+     * We would like to wait for a total of ~4.5 seconds to account for slight drift/jitter.
+     * We have been waiting for 0.8*4=3.2 sec. Wait for another 1 sec here.
+     */
+    aws_thread_current_sleep(1000000000);
```
bretambrose, line 3503: I don't see any asserts in this test actually checking anything meaningful.

xiazhvera:
Updated the unit tests.

tests/v3/connection_state_test.c
```diff
@@ -3421,6 +3433,74 @@ AWS_TEST_CASE_FIXTURE(
...
+
+    uint64_t begin_timestamp = 0;
+    uint64_t elapsed_time = 0;
+    uint64_t now = 0;
+    aws_high_res_clock_get_ticks(&begin_timestamp);
+    uint64_t test_duration = (uint64_t)4 * AWS_TIMESTAMP_NANOS;
+
+    // Make sure we publish for 4 seconds;
+    while (elapsed_time < test_duration) {
+        /* Publish qos1*/
```
bretambrose, line 3469: qos0

bretambrose (approved):
Fix comment and ship

PR #317, opened 2023-08-15
Add mqtt311 on_termination callback

sfod:
*Issue #, if available:*

*Description of changes:*


include/aws/mqtt/client.h
```diff
@@ -161,6 +161,8 @@ typedef void(aws_mqtt_client_publish_received_fn)(
 /** Called when a connection is closed, right before any resources are deleted */
 typedef void(aws_mqtt_client_on_disconnect_fn)(struct aws_mqtt_client_connection *connection, void *userdata);
 
+typedef void(aws_mqtt_client_on_termination_fn)(struct aws_mqtt_client_connection *connection, void *userdata);
```
xiazhvera, line 164: The termination callback probably should not take the `connection` as a parameter, as the connection might be destroyed or at least would be unaccessible soon.

source/client.c
```diff
@@ -890,6 +899,10 @@ static void s_mqtt_client_connection_start_destroy(struct aws_mqtt_client_connec
     if (call_destroy_final) {
         s_mqtt_client_connection_destroy_final(&connection->base);
     }
+
+    if (client_termination_handler != NULL) {
```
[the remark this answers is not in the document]
sfod: You're right, now I see that it should be called from s_mqtt_client_connection_destroy_final.

tests/v3/connection_state_test.c
```diff
@@ -3682,3 +3685,146 @@ AWS_TEST_CASE_FIXTURE(
...
+    struct mqtt_connection_state_test *state_test_data = (struct mqtt_connection_state_test *)userdata;
+
+    aws_mutex_lock(&state_test_data->lock);
+    state_test_data->connection_termination_calls += 1;
+    aws_mutex_unlock(&state_test_data->lock);
+
+    state_test_data->connection_terminated = true;
+}
+
+static int s_clean_up_mqtt_server_with_termination_fn(struct aws_allocator *allocator, int setup_result, void *ctx) {
```
xiazhvera, line 3712: Could we just add the termination change directly to `s_clean_up_mqtt_server_fn`? In that case we could have termination callback attached to all unit tests.

sfod: Initially I did exactly that. But then I thought that separating tested functionality as much as possible would be better. Otherwise, if for some reason `on_termination` callback breaks, it'll cause all other tests in CI to fail, making it more difficult to find the cause of the issue. With the current approach we'll see a couple `test_on_termination_callback_*` failed, and it'll be easy enough to understand where to look.
Saying that, I see that enabling `on_termination` callback for all tests may cover additional cases that I missed or that will appear in the future.

xiazhvera: I think we can have both. A separate simply for `on_termination` and also applied it to all other tests. If the separate one failed, that's good enough for us to understand that the other tests are also failing because of the `on_termination`.

sfod: That sounds reasonable, I'll enable on_termination for all tests

tests/v3/connection_state_test.c
```diff
@@ -266,6 +269,29 @@ static void s_wait_for_connection_to_fail(struct mqtt_connection_state_test *sta
...
+}
+
+static void s_on_connection_termination_fn(void *userdata) {
+    struct mqtt_connection_state_test *state_test_data = (struct mqtt_connection_state_test *)userdata;
+
+    aws_mutex_lock(&state_test_data->lock);
+    state_test_data->connection_termination_calls += 1;
+    aws_mutex_unlock(&state_test_data->lock);
+
+    state_test_data->connection_terminated = true;
```
xiazhvera, line 292: The `connection_terminated` should set with in the data lock.

bretambrose (approved):

PR #319, opened 2023-08-16
Add on_termination callback for adapter

xiazhvera:
*Issue #, if available:*

*Description of changes:*


source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -69,6 +69,13 @@ static void s_mqtt_adapter_final_destroy_task_fn(struct aws_task *task, void *ar
 
     AWS_LOGF_DEBUG(AWS_LS_MQTT5_TO_MQTT3_ADAPTER, "id=%p: Final destruction of mqtt3-to-5 adapter", (void *)adapter);
 
+    /* trigger the termination callback */
```
bretambrose, line 72: As a general pattern, termination handlers should be invoked as the last possible thing, after memory has been deallocated (which means you need to save off the pointers earlier before freeing).   Cleaning up the operation table might generate callbacks.  Tests that wait for the callback may expect all memory to be released.  etc...

bretambrose (approved):

PR #321, opened 2023-08-22
Validation phase of MQTT311 compliance pass

bretambrose:
A set of validation-related updates for MQTT311 compliance.

* subscribe_multiple is validated for at least one subscription
* unsubscribe topic must be non-null
* publish now rejects invalid qos values
* additional tests for topic filter and topic validation

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -42,6 +42,7 @@ add_test_case(mqtt_topic_tree_unsubscribe)
 add_test_case(mqtt_topic_tree_duplicate_transactions)
 add_test_case(mqtt_topic_tree_transactions)
 add_test_case(mqtt_topic_validation)
+add_test_case(mqtt_topic_filter_validation)
 
 add_test_case(mqtt_connect_disconnect)
 add_test_case(mqtt_connect_set_will_login)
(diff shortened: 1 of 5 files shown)


tests/CMakeLists.txt
```diff
@@ -86,6 +87,10 @@ add_test_case(mqtt_connection_reconnection_backoff_unstable)
 add_test_case(mqtt_connection_reconnection_backoff_reset)
 add_test_case(mqtt_connection_reconnection_backoff_reset_after_disconnection)
 
+add_test_case(mqtt_validation_failure_publish_qos)
+add_test_case(mqtt_validation_failure_subscribe_empty)
+add_test_case(mqtt_validation_failure_unsubscribe_null)
```
xiazhvera, line 92: The added the unit test is testing for NULL topic, while we added the validation for NULL payload. Should we have a test for the null payload as well?

bretambrose: Null payloads are fine so it's not really a validation test.  We could theoretically add a null payload test to standard client tests and verify it works as expected.  I'll add that when I get to client behavior.

xiazhvera (approved):

tests/v3/topic_tree_test.c
```diff
@@ -316,18 +316,51 @@ static int s_mqtt_topic_validation_fn(struct aws_allocator *allocator, void *ctx
...
+
+    ASSERT_TOPIC_VALIDITY(FALSE, "#");
+    ASSERT_TOPIC_VALIDITY(FALSE, "sport/tennis/#");
     ASSERT_TOPIC_VALIDITY(FALSE, "sport/tennis#");
     ASSERT_TOPIC_VALIDITY(FALSE, "sport/tennis/#/ranking");
-
-    ASSERT_TOPIC_VALIDITY(TRUE, "+");
-    ASSERT_TOPIC_VALIDITY(TRUE, "+/tennis/#");
-    ASSERT_TOPIC_VALIDITY(TRUE, "sport/+/player1");
+    ASSERT_TOPIC_VALIDITY(FALSE, "");
```
sbSteveK, line 331: Probably unnecessary but we're not checking a NULL pointer. `s_is_valid_topic()` returns false on`topic == NULL`.

sbSteveK (approved):

PR #327, opened 2023-09-18
Fix: coredump, dereference null variable

alfred2g:


bretambrose (commented):
Can we add a test that triggers an error-code based completion?

Probably the simplest way of doing that is to set a low operation timeout and arrange the mqtt5 server mock to not response to subscribes at all.  The operation will time out and there won't be a suback.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2184,10 +2184,14 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
     void *complete_ctx) {
-    (void)suback;
 
     struct aws_mqtt5_to_mqtt3_adapter_operation_subscribe *subscribe_op = complete_ctx;
     struct aws_mqtt_client_connection_5_impl *adapter = subscribe_op->base.adapter;
+    size_t reason_code_count = 0;
+
+    if (suback != NULL) {
+        reason_code_count = suback->reason_code_count;
+    }
```
xiazhvera, line 2194: Probably move this down to the `granted_qos` part, so we only need to do the check once. 
```
if (suback == NULL) {
            granted_qos = AWS_MQTT_QOS_FAILURE;
        }
```

tests/v5/mqtt5_to_mqtt3_adapter_tests.c
```diff
@@ -2686,6 +2686,91 @@ static int s_mqtt5to3_adapter_subscribe_single_success_fn(struct aws_allocator *
 
 AWS_TEST_CASE(mqtt5to3_adapter_subscribe_single_success, s_mqtt5to3_adapter_subscribe_single_success_fn)
 
+/*
+ * This function tests receiving a subscribe acknowlege after disconnecting from
+ * the server.
+ * it expects a AWS_MQTT_QOS_FAILURE return
+ */
+static int s_mqtt5to3_adapter_subscribe_single_success_null_suback_fn(struct aws_allocator *allocator, void *ctx) {
```
xiazhvera, line 2694: Probably remove the `success` from the name. It is not testing a success case here.

tests/v5/mqtt5_to_mqtt3_adapter_tests.c
```diff
@@ -2686,6 +2686,91 @@ static int s_mqtt5to3_adapter_subscribe_single_success_fn(struct aws_allocator *
...
+    (void)ctx;
+
+    aws_mqtt_library_init(allocator);
+
+    struct mqtt5_client_test_options test_options;
+    aws_mqtt5_client_test_init_default_options(&test_options);
+
+    test_options.server_function_table.packet_handlers[AWS_MQTT5_PT_SUBSCRIBE] =
+        s_mqtt5_mock_server_handle_subscribe_suback_success;
```
xiazhvera, line 2703: Should we set this to NULL so that we guarantee that the server never respond with a `suback`?

source/v5/mqtt5_to_mqtt3_adapter.c
alfred2g: Added it twice so i don't change the scope of the variable granted_qos

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2226,23 +2234,22 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
+            struct aws_mqtt_topic_subscription *, multi_sub_subscription_ptr_buf, reason_code_count);
         struct aws_mqtt_topic_subscription *subscription_ptr =
             (struct aws_mqtt_topic_subscription *)multi_sub_subscription_buf;
 
         struct aws_array_list multi_sub_list;
         aws_array_list_init_static(
             &multi_sub_list,
             multi_sub_subscription_ptr_buf,
-            suback->reason_code_count,
+            reason_code_count,
```
sfod, line 2247: `aws_array_list_init_static` has a check for the third parameter:
```c
    AWS_FATAL_PRECONDITION(item_count > 0);
```

It seems we should handle `reason_code_count == 0` separately.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2207,7 +2215,7 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
             topic_filter = record->subscription_view.topic_filter;
         }
 
-        if (suback->reason_code_count > 0) {
+        if (reason_code_count > 0) {
             granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
         }
```
sbSteveK, line 2220: Trivial: Instead of checking on L2206, could we add an else block and assign `AWS_MQTT_QOS_FAILURE` to `granted_qos` here to simplify where that is getting assigned to one place? It'd also remove the second `suback == NULL` check that Vera mentioned.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2226,23 +2234,22 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
             "id=%p: mqtt3-to-5-adapter, completing multi-topic subscribe",
             (void *)adapter);
 
+        AWS_VARIABLE_LENGTH_ARRAY(struct aws_mqtt_topic_subscription, multi_sub_subscription_buf, reason_code_count);
```
sfod, line 2237: With `reason_code_count` equal to 0, the expression here will be
```c
struct aws_mqtt_topic_subscription multi_sub_subscription_buf[0];
```
, which is invalid.

It's just another reason to handle `reason_code_count == 0` separately.

source/v5/mqtt5_to_mqtt3_adapter.c
sbSteveK: It sounds like a null suback may be normal and expected behavior. I don't think we want to fail the `AWS_FATAL_PRECONDITION` on something we're doing intentionally (e.g. op timeout, close connection). We may want to set the `subscription_count` earlier and use it since it's grabbing that from the stored `&subscribe_op->subscriptions)` array length on L2250.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2226,23 +2234,22 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
             &multi_sub_list,
             multi_sub_subscription_ptr_buf,
-            suback->reason_code_count,
+            reason_code_count,
             sizeof(struct aws_mqtt_topic_subscription *));
 
         size_t subscription_count = aws_array_list_length(&subscribe_op->subscriptions);
 
-        for (size_t i = 0; i < suback->reason_code_count; ++i) {
+        for (size_t i = 0; i < reason_code_count; ++i) {
```
sbSteveK, line 2252: same here as above. Use subscription_count to process and report back the failed sub for all subscriptions. We will also need to assign the `subscription->qos` to the correct response.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2226,23 +2234,22 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
             struct aws_mqtt_topic_subscription *subscription = subscription_ptr + i;
             AWS_ZERO_STRUCT(*subscription);
```
sbSteveK, line 2255: Check `reason_code_count` or `suback == NULL` so we don't try to access `suback->reason_codes` if it's NULL and set the `subscription->qos` to`AWS_MQTT_QOS_FAILURE`

source/v5/mqtt5_to_mqtt3_adapter.c
alfred2g: can't we have a reason code of zero without suback being null?

source/v5/mqtt5_to_mqtt3_adapter.c
alfred2g: it is an expected behavior, this function is called with an explicit null as the suback

xiazhvera: Looking at this case, we would need a test for suback on multiple subscription as well.

source/v5/mqtt5_to_mqtt3_adapter.c
sfod: According to [the docs](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901178), the number of reason codes must match the number of topics. Though, I'd say a simple check probably won't hurt.

source/v5/mqtt5_to_mqtt3_adapter.c
sbSteveK: Ah right, maybe we need a test that sets the multi suback but not the singular one so it hits the multi processing.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2184,89 +2184,109 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
+                error_code,
+                subscribe_op->on_suback_user_data);
+        }
+        if (subscribe_op->on_multi_suback != NULL) {
+            AWS_LOGF_DEBUG(
+                AWS_LS_MQTT5_TO_MQTT3_ADAPTER,
+                "id=%p: mqtt3-to-5-adapter, completing multi-topic subscribe",
+                (void *)adapter);
+            (*subscribe_op->on_multi_suback)(
+                &adapter->base, subscribe_op->base.id, NULL, error_code, subscribe_op->on_multi_suback_user_data);
```
sfod, line 2211: `aws-crt-cpp` expects the list to be non-empty. It won't crash, but it'll set the qos to `AWS_MQTT_QOS_AT_MOST_ONCE` and then pass empty list of topics to the user callback. Which may be confusing.
Here's the corresponding code: https://github.com/awslabs/aws-crt-cpp/blob/main/source/mqtt/MqttConnectionCore.cpp#L414

Probably the empty list should be handled in `aws-crt-cpp`. But I feel like it should be done here, so all the layers above don't have to worry about these things.

xiazhvera: I thought about it. I think it might be better to handle it on CRT side, as it might be better to check NULL value for a pointer regardless. 
I will update the changes in CRTs.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2197,20 +2197,23 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
 
         struct aws_byte_cursor topic_filter;
         AWS_ZERO_STRUCT(topic_filter);
+
         enum aws_mqtt_qos granted_qos = AWS_MQTT_QOS_AT_MOST_ONCE;
 
+        aws_array_list_get_at(&subscribe_op->subscriptions, &record, 0);
```
sfod, line 2203: Getting the record duplicated here from the `if` block below, looks suspicious

sfod (approved):

tests/v5/mqtt5_to_mqtt3_adapter_tests.c
```diff
@@ -2814,6 +2896,88 @@ static int s_mqtt5to3_adapter_subscribe_multi_success_fn(struct aws_allocator *a
...
+static int s_mqtt5to3_adapter_subscribe_multi_null_suback_fn(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+
+    aws_mqtt_library_init(allocator);
+
+    struct mqtt5_client_test_options test_options;
+    aws_mqtt5_client_test_init_default_options(&test_options);
+
+    test_options.server_function_table.packet_handlers[AWS_MQTT5_PT_SUBSCRIBE] =
+        s_mqtt5_mock_server_handle_subscribe_suback_success;
```
xiazhvera, line 2913: The server should not send suback back.

PR #328, opened 2023-09-22
Sync up on_suback_multi behavior between adapter and Mqtt3 client

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/v5/mqtt5_to_mqtt3_adapter.c
+++ b/source/v5/mqtt5_to_mqtt3_adapter.c
@@ -2188,6 +2188,12 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
     struct aws_mqtt5_to_mqtt3_adapter_operation_subscribe *subscribe_op = complete_ctx;
     struct aws_mqtt_client_connection_5_impl *adapter = subscribe_op->base.adapter;
     struct aws_mqtt_subscription_set_subscription_record *record = NULL;
+    size_t reason_code_count = 0;
+
+    if (suback != NULL) {
+        reason_code_count = suback->reason_code_count;
+    }
+    size_t subscription_count = aws_array_list_length(&subscribe_op->subscriptions);
 
     if (subscribe_op->on_suback != NULL) {
         AWS_LOGF_DEBUG(
@@ -2200,16 +2206,15 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
 
         enum aws_mqtt_qos granted_qos = AWS_MQTT_QOS_AT_MOST_ONCE;
 
-        size_t subscription_count = aws_array_list_length(&subscribe_op->subscriptions);
         if (subscription_count > 0) {
             aws_array_list_get_at(&subscribe_op->subscriptions, &record, 0);
             topic_filter = record->subscription_view.topic_filter;
         }
 
-        if (suback != NULL) {
-            if (suback->reason_code_count > 0) {
-                granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
-            }
+        if (reason_code_count > 0) {
+            granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
+        } else if (record) {
+            granted_qos = (enum aws_mqtt_qos)(record->subscription_view.qos);
         } else {
             granted_qos = AWS_MQTT_QOS_FAILURE;
         }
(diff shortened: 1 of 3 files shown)


source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2228,50 +2233,51 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
+
+                subscription->topic = record->subscription_view.topic_filter;
+                subscription->on_publish = record->subscription_view.on_publish_received;
+                subscription->on_publish_ud = record->subscription_view.callback_user_data;
+                subscription->on_cleanup = record->subscription_view.on_cleanup;
+            }
 
-                aws_array_list_push_back(&multi_sub_list, &subscription);
+            if (suback != NULL) {
+                subscription->qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[i]);
```
sfod, line 2268: Debatable: If `suback` is not null then we received it from the server, and according to the mqtt standard the size of `suback->reason_codes` must be equal to the number of subscriptions. However, on the line 2258 we handle the case `length(suback->reason_codes) > subscription_count`. For consistency, we should handle here the case `suback != null && length(suback->reason_codes) == 0`.
Or we can verify that `length(suback->reason_codes) == subscription_count` prior to this loop. And if it's not true, idk, at least add error log.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2228,50 +2233,51 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
-
-            for (size_t i = 0; i < suback->reason_code_count; ++i) {
-                struct aws_mqtt_topic_subscription *subscription = subscription_ptr + i;
-                AWS_ZERO_STRUCT(*subscription);
-
-                subscription->qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[i]);
-
-                if (i < subscription_count) {
-                    aws_array_list_get_at(&subscribe_op->subscriptions, &record, i);
+        // If the suback does not contains any data, we directly extract the data from subscribe_op
```
sbSteveK, line 2236: Trivial: "If suback does not cotain data, we extract data from subscribe_op"

source/v5/mqtt5_to_mqtt3_adapter.c
sbSteveK: Debatable: I think instead of setting `reason_code_count = subscription_count` in the event of `reason_code_count == 0`, we should just use `subscription_count` in the creation of arrays and iterating in this loop. Then we could use `if (i < reason_code_count)` here which would account for both a null suback and `length(suback->reason_codes) ==0` as well as being able to remove the `if (i < subscription_count)` check on 2258 since the loop is already doing that check.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2200,16 +2206,15 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
             topic_filter = record->subscription_view.topic_filter;
         }
 
-        if (suback != NULL) {
-            if (suback->reason_code_count > 0) {
-                granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
-            }
+        if (reason_code_count > 0) {
+            granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
+        } else if (record) {
```
bretambrose, line 2216: I am confused as to what's going on here.  Under no circumstances should we ever "invent" reason codes where none exist.  A suback with no reason codes in it (or a mismatch relative to the subscribe) should tear down the connection as a protocol error.  If we're not doing so atm, we should.

bretambrose (changes requested):

source/v5/mqtt5_to_mqtt3_adapter.c
sbSteveK: MQTT3 on a timeout will return a timeout error code with the original subscribe request topic and request qos. If we have no suback packet due to a timeout or shutting down the client in mqtt5, to emulate mqtt3 behavior we need to also send back the same along with the error code.

sbSteveK: haha, yea. I had the same thought when I checked. The sins of our fathers...

bretambrose: Do any bindings do anything with those values when an error is set?  There are (buggy) cases where an error may not get set properly in which case doing this makes it look like the subscribe succeeded.

sfod: Yep, aws-crt-cpp extracts info regardless of the error code: https://github.com/awslabs/aws-crt-cpp/blob/main/source/mqtt/MqttConnectionCore.cpp#L417

sbSteveK: Python bindings returns a future on a subscribe() and if there's an error code, sets an exception from the error_code in the future...

sbSteveK: Yikes. So in cpp, it's up to the customer to check the error code to make sure they're not assuming successful subscribes. I need to double-check that we're not at least setting the request.qos to fail in a timeout. If we are, then we should be fine everywhere

source/v5/mqtt5_to_mqtt3_adapter.c
xiazhvera: > Debatable: If `suback != null && length(suback->reason_codes) == 0`, we'll get out of bounds access here.
> 
> It's debatable becuase when `suback` is not null then we received it from the server, and according to the mqtt standard the size of `suback->reason_codes` must be equal to the number of subscriptions. However, on the line 2258 we handle the case `length(suback->reason_codes) > subscription_count`, i.e. we handle the situation when the received `suback` packet violates the standard. So, for consistency, we should handle here the case `suback != null && length(suback->reason_codes) == 0`. Or, maybe it'll be better to verify that `length(suback->reason_codes) == subscription_count` prior to this loop. And if it's not true, idk, at least add error log.

We don't need to worry about out of bounds here because the loop will loop through the `reason_code_count` as long as `suback!=NULL`. If `suback->reason_codes) == 0`, we will not go into the loop here.

source/v5/mqtt5_to_mqtt3_adapter.c
xiazhvera: Java binding will directly return error_code, while JS binding will extract the data regardless of the error_code. 

It is probably reasonable to keep the topic and qos data when the subscription failed with an error_code. The user still wanna get the information of a failing subscription for possible operations. (For a simply example, the user would need the info to print out a proper error message, or try subscribe again on the same topic/qos value.)

xiazhvera:
I leave the `on_suback` behavior as it was in origin Mqtt311, where the client would return the subscription data even for a failure subscription. I think the user still wanna get the information of a failing subscription for possible operations. (For example, the user would need the info to print out a proper error message, or try subscribe again on the same topic/qos value)

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2198,21 +2200,19 @@ static void s_aws_mqtt5_to_mqtt3_adapter_subscribe_completion_fn(
...
-        if (suback != NULL) {
-            if (suback->reason_code_count > 0) {
-                granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
-            }
-        } else {
-            granted_qos = AWS_MQTT_QOS_FAILURE;
+        if (suback != NULL && suback->reason_code_count > 0) {
+            granted_qos = s_convert_mqtt5_suback_reason_code_to_mqtt3_granted_qos(suback->reason_codes[0]);
+        } else if (record) {
+            granted_qos = (enum aws_mqtt_qos)(record->subscription_view.qos);
```
bretambrose, line 2213: Looking at this again, I think this is not what we want to do.  This is a bug in the 311 client (to report the requested qos rather than the returned qos).  We should fix the bug instead of making the 5 adapater also do the wrong thing.

This would apply to both the single and multi acks.

xiazhvera:
Fixed mqtt3 behavior https://github.com/awslabs/aws-c-mqtt/pull/340 instead of updating the adapter.

PR #331, opened 2023-10-20
Fail operations when adapter is in destroy process

xiazhvera:
*Issue #, if available:*

*Description of changes:*


include/aws/mqtt/private/v5/mqtt5_to_mqtt3_adapter_impl.h
```diff
@@ -257,6 +257,12 @@ struct aws_mqtt_client_connection_5_impl {
      */
     struct aws_ref_count internal_refs;
 
+    /*
+     * Set to true if the adapter is in destroying process (internal_ref is 0). The adapter would prevent any
+     * operations if `is_destroying` set to true
+     */
+    bool is_destroying;
```
xiazhvera, line 264: Judgement call: Instead of creating a new member here, we can use `adapter_state` to track the destroy processing. When we start the destroy process (internal ref -> 0), the adapter would not receive any lifecycle events anymore.

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -327,6 +344,10 @@ static int s_aws_mqtt_client_connection_5_connect(
     const struct aws_mqtt_connection_options *connection_options) {
 
     struct aws_mqtt_client_connection_5_impl *adapter = impl;
+    int result = s_check_adapter_destroy_status(adapter);
```
sfod, line 347: This check is missing for `s_aws_mqtt_5_resubscribe_existing_topics`, `s_aws_mqtt_client_connection_5_acquire` and `s_aws_mqtt_client_connection_5_release`. I thing, that's all but I could something

xiazhvera: Updated `s_aws_mqtt_5_resubscribe_existing_topics`.

Debatable: I feel we should directly fail the `acquire`/`release` if the adapter is in destroying process... the ASSERT instead of a validation check might be a better option here.

sfod: Yeah, this situation (trying to do something meaningful with `connection` in callback while it's being destroyed in another thread) should be pretty rare and most probably will indicate that something's wrong with user code. So, ASSERT could actually help to find a possible issue in the code.

sfod:
Debatable: I feel slightly uneasy because of passing a "half-dead" `connection` object into callbacks. But this approach seems safe with the current implementation, so I don't have strong opinion about it.

bretambrose (changes requested):
Usage after ref-count goes to zero is not something we need to worry about.

xiazhvera:
Talked to Bret about the issue. The caller should be responsible to track the `connection` reference. If the caller have released the connection reference, the caller should never use `connection` even if the `connection` is from the callback.

As Igor mentioned, passing a "half-dead" connection object into callbacks is hard to handle. However, if we pass a null here, it has a risk to crash the user. (If the caller forget to handle NULL here)

We should make a rule to avoid expose the object in callbacks in the future implementation.

PR #333, opened 2023-10-27
Mqtt5 operation timeout

bretambrose:
* Adds support to override the ack timeout on a per-operation basis in the MQTT5 client


tests/v5/mqtt5_client_tests.c
```diff
@@ -6076,3 +6076,299 @@ static int s_mqtt5_client_outbound_alias_lru_success_a_b_c_br_cr_a_fn(struct aws
...
+ * unsubscribe with 4 second timeout
+ * qos 1 publish with 1 second timeout
+ *
+ * We expect to see callbacks in sequence:
+ *
+ *  qos 0 publish success
+ *  qos 1 publish failure by timeout after 1 second
+ *  subscribe failure by timeout after 2 seconds
+ *  qos 1 publish failure by timeout after 3 seconds
+ *  unsubscribe failure by timeout after 4 seconds
```
sfod, line 6211: Debatable: I feel that another operation with timeout set to 0 would be good for coverage, as we process 0 differently. But I don't see a way to add this easily. So, I just wanted to point it out, maybe you have an idea how it can be covered without complicating things.

sfod (approved):

tests/v5/mqtt5_client_tests.c
bretambrose: We could add a test where the client has a default and we submit with 0 and assert that it uses the default.

PR #334, opened 2023-11-09
Relax validation on user-specified topic aliasing

bretambrose:
Being strict about topic alias validation seems counter productive.  It's an optional micro-optimization and the client configuration may not be known to a user who may have to work with a client configured by a third party.


source/v5/mqtt5_encoder.c
```diff
@@ -800,7 +800,25 @@ static int s_aws_mqtt5_encoder_begin_publish(struct aws_mqtt5_encoder *encoder,
...
+                (void *)encoder->config.client,
+                outbound_topic_alias);
+            if (outbound_topic.len == 0) {
+                AWS_LOGF_DEBUG(
+                    AWS_LS_MQTT5_GENERAL,
+                    "(%p) mqtt5 client encoder - PUBLISH packet dropping topic field for previously established alias",
+                    (void *)encoder->config.client);
+            }
             local_publish_view.topic_alias = &outbound_topic_alias;
+        } else {
```
sbSteveK, line 815: Trivial: We could probably do without this log. I may be missing something but I don't think it's providing any useful information and the absence of the topic alias log above should be enough to infer that the publish isn't using topic aliasing. Also unsure whether we need to NULL

I think local_publish_view.topic_alias = NULL may be unnecessary. if outbound_topic_alias == 0, it's already NULL or an error was raised by the resolver. Could be missing something deeper tho.

source/v5/mqtt5_options_storage.c
```diff
@@ -1743,30 +1743,6 @@ static int s_aws_mqtt5_packet_publish_view_validate_vs_connection_settings(
             return aws_raise_error(AWS_ERROR_MQTT5_PUBLISH_OPTIONS_VALIDATION);
         }
 
-        if (publish_view->topic_alias != NULL) {
-            const struct aws_mqtt5_client_options_storage *client_options = client->config;
-            if (client_options->topic_aliasing_options.outbound_topic_alias_behavior != AWS_MQTT5_COTABT_USER) {
```
sbSteveK, line 1748: Trivial: Not erroring here seems fine but should we log that we're ignoring?

sbSteveK (approved):
Trivial comments on logging

source/v5/mqtt5_encoder.c
bretambrose: null assignment is necessary; up until now we're using what is specified by the user which could be an invalid mapping.

source/v5/mqtt5_options_storage.c
bretambrose: I feel like encoder logging gives a definitive answer to what we do alias-wise.

source/v5/mqtt5_encoder.c
bretambrose: As an addendum, we're always in a better position if the logging tells us exactly what happened on its own, rather than having to read the log and the code and infer what happened based on what didn't show up in the logs.

PR #340, opened 2024-01-09
Fix mqtt3 suback returned qos behavior

xiazhvera:
*Issue #, if available:*

*Description of changes:*
The Mqtt3 client would now return `AWS_MQTT_QOS_FAILURE` as qos value if the subscription failed with error code. 
While Mqtt3 client would directly return the request qos value. This change would introduce behavior changes in CPP, Python and NodeJS. 


source/client.c
```diff
@@ -1945,19 +1945,23 @@ static void s_subscribe_complete(
...
+            if (error_code != AWS_OP_SUCCESS) {
+                subscription->qos = AWS_MQTT_QOS_FAILURE;
+            }
             err |= aws_array_list_push_back(&cb_list, &subscription);
         }
         AWS_ASSUME(!err);
         task_arg->on_suback.multi(&connection->base, packet_id, &cb_list, error_code, task_arg->on_suback_ud);
         aws_array_list_clean_up(&cb_list);
     } else if (task_arg->on_suback.single) {
+        // The topic->request.qos should be already updated to returned qos
```
sbSteveK, line 1958: Trivial: Here and below in `s_subscribe_single_complete` we could probably change the topic->request-qos directly instead of creating a separate `returned_qos`. I don't believe the topic->request is used after the point of invoking the `on_suback` callbacks in either case.

sbSteveK (approved):

tests/v3/connection_state_test.c
```diff
@@ -1614,6 +1694,86 @@ AWS_TEST_CASE_FIXTURE(
     s_clean_up_mqtt_server_fn,
     &test_data)
 
+/* Subscribe to multiple topics prior to connection, make a CONNECT, have the server send PUBLISH messages,
+ * make sure they're received, then send a DISCONNECT. */
```
sfod, line 1698: Trivial: The test description seems to be incorrect

xiazhvera:
Hold off because it introduce a behavior change.

PR #342, opened 2024-01-19
Protocol adapter skeleton

bretambrose:
* Request-response protocol adapter interface
* MQTT5 specific implementation of the request-response protocol adapter interface


include/aws/mqtt/private/client_impl_shared.h
```diff
@@ -10,6 +10,20 @@
...
 
+/*
+ * Internal enum that indicates what type of struct the underlying impl pointer actually is.  We use this
+ * to safely interact with private APIs on the implementation or extract the adapted 5 client directly, as
+ * necessary.
+ */
+enum aws_mqtt311_impl_type {
+
+    /* 311 connection impl can be cast to `struct aws_mqtt_client_connection_311_impl` */
+    AWS_MQTT311_IT_311_CONNECTION_IMPL,
```
sbSteveK, line 21: Trivial/Debatable: Can we remove the "IMPL" from the backs or remove "IT" from the mid of these enums? I think it'd make more sense to remove the "IMPL" at the end to follow the `aws_mqtt311_impl_type` name. e.g. "AWS_MQTT311_IT_311_CONNECTION"

include/aws/mqtt/private/request-response/protocol_adapter.h
```diff
@@ -0,0 +1,196 @@
...
+    const struct aws_mqtt_protocol_adapter_vtable *vtable;
+    void *impl;
+};
+
+AWS_EXTERN_C_BEGIN
+
+/*
+ * Creates a new request-response protocol adapter from an MQTT311 client
+ */
+AWS_MQTT_API struct aws_mqtt_protocol_adapter *aws_mqtt_protocol_adapter_new_from_311(
```
sbSteveK, line 154: Trivial: We usually stick to the format of `_new()` and pair it with `_destroy()`. Just wondering if this is different for any particular reason. e.g. why aren't we using `aws_mqtt_protocol_adapter_from_311_new()` and `aws_mqtt_protocol_adapter_destroy()`?

include/aws/mqtt/private/request-response/protocol_adapter.h
```diff
@@ -0,0 +1,196 @@
...
+#include <aws/common/byte_buf.h>
+
+struct aws_allocator;
+struct aws_mqtt_client_connection;
+struct aws_mqtt5_client;
+
+/*
+ * The request-response protocol adapter is a translation layer that sits between the request-response native client
+ * implementation and a protocol client capable of subscribing, unsubscribing, and publishing MQTT messages.
+ * Valid protocol clients include the CRT MQTT5 client, the CRT MQTT311 client, and an eventstream RPC connection
```
sbSteveK, line 21: Trivial: Should we exclude eventstream RPC from the description until it's implemented?

sbSteveK:
We might want to update mqtt5_client.h "@param completion_options" to include a short description of the recently added `ack_timeout_seconds_override` members since the options aren't really addressed anywhere else. We could also include somewhere in the code comments or some documentation around  the request-response various option structs what's up with the `ack_timeout_seconds` each has and what its default value does. Mostly for ourselves in case we're looking.

include/aws/mqtt/private/request-response/protocol_adapter.h
bretambrose: Pattern is <subject><verb><qualifiers>

For ex: https://github.com/awslabs/aws-c-auth/blob/main/include/aws/auth/credentials.h#L650-L1116

include/aws/mqtt/private/request-response/protocol_adapter.h
bretambrose: I don't see the need.  This is in a private and will get moved to a different repository eventually anyways.

include/aws/mqtt/private/request-response/protocol_adapter.h
sbSteveK: Ah, gotcha. I don't think I've seen the qualifiers around till now.

source/request-response/protocol_adapter.c
```diff
@@ -0,0 +1,422 @@
...
+    .aws_mqtt_protocol_adapter_unsubscribe_fn = s_aws_mqtt_protocol_adapter_5_unsubscribe,
+    .aws_mqtt_protocol_adapter_publish_fn = s_aws_mqtt_protocol_adapter_5_publish,
+};
+
+struct aws_mqtt_protocol_adapter *aws_mqtt_protocol_adapter_new_from_5(
+    struct aws_allocator *allocator,
+    struct aws_mqtt_protocol_adapter_options *options,
+    struct aws_mqtt5_client *client) {
+    struct aws_mqtt_protocol_adapter_5_impl *adapter =
+        aws_mem_calloc(allocator, 1, sizeof(struct aws_mqtt_protocol_adapter_5_impl));
```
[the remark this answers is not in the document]
bretambrose: This isn't a public API though.

source/request-response/weak_ref.c
```diff
@@ -0,0 +1,54 @@
...
+
+#include <aws/common/ref_count.h>
+
+struct aws_weak_ref {
+    struct aws_allocator *allocator;
+    struct aws_ref_count refcount;
+    void *referenced;
+};
+
+static void s_destroy_weak_ref(void *value) {
```
sbSteveK, line 16: nitpick/trivial: structure of [subject][verb][qualifiers] in `destroy` being at the end of most of these calls. I found we use a lot of `s_destroy_[subject]` in our mqtt5 implementation so maybe this is an alternate pattern we use so we can keep as is?

source/request-response/weak_ref.c
```diff
@@ -0,0 +1,54 @@
...
+struct aws_weak_ref *aws_weak_ref_release(struct aws_weak_ref *weak_ref) {
+    if (NULL != weak_ref) {
+        aws_ref_count_release(&weak_ref->refcount);
+    }
+
+    return NULL;
+}
+
+void *aws_weak_ref_get_reference(struct aws_weak_ref *weak_ref) {
+    return weak_ref->referenced;
```
sbSteveK, line 49: Do we need a NULL check on the pointer here before returning the referenced?

source/request-response/protocol_adapter.c
```diff
@@ -0,0 +1,428 @@
...
+
+/*
+ * New API contract
+ *
+ * Invariant 1: Subscribe is only called from the RR subscription manager when going from 0 to 1 pending operations
+ * Invariant 2: Unsubscribe is only called from the RR subscription manager when there are 0 pending operations, not
+ *   necessarily on the exact transition to zero though.
+ *
+ * Entries are not tracked with the exception of eventstream impl which needs the stream handles to close.
+ * A subscribe failure should not trigger an unsubscribe, only notify the status callback.
```
sbSteveK, line 23: Should we include in here that we also don't retry a subscription? "A subscribe failure should not trigger an unsubscribe or retry a subscription, only notify the status callback." ?

source/request-response/protocol_adapter.c
```diff
@@ -0,0 +1,428 @@
...
+ * Subscription event callback should be {subscribe_success, subscribe_failure, unsubscribe_success,
+ * unsubscribe_failure}. The sub manager is responsible for calling Unsubscribe on all its entries when shutting down
+ * (before releasing hold of the adapter).
```
sbSteveK, line 26: debatable/trivial: maybe move the sentence about the sub manager unsubscribes during shutdown to the end of this api contract description block?

source/request-response/weak_ref.c
bretambrose: The names of internal static functions aren't really critical

source/request-response/weak_ref.c
bretambrose: I don't see why; if it's null we should be crashing

source/request-response/protocol_adapter.c
bretambrose: This is just me talking to myself.  If it creates confusion I'll just remove.

source/request-response/protocol_adapter.c
```diff
@@ -0,0 +1,428 @@
...
+    struct aws_event_loop *loop;
+    struct aws_mqtt5_client *client;
+    struct aws_mqtt5_listener *listener;
+};
+
+static void s_aws_mqtt_protocol_adapter_5_destroy(void *impl) {
+    struct aws_mqtt_protocol_adapter_5_impl *adapter = impl;
+
+    // all the real cleanup is done in the listener termination callback
+    aws_mqtt5_listener_release(adapter->listener);
```
sbSteveK, line 65: When the last reference is dropped to the protocol adapter, it calls this destructor, which drops its reference to the listener and triggers the listener's cleanup/destruction which then checks its config and calls into `s_protocol_adapter_mqtt5_listener_termination_callback`. This feels like unnecessary complexity. Is there a reason why we aren't executing what's happening in `s_protocol_adapter_mqtt5_listener_termination_callback` in this destroy function after releasing the listener? It'd be easier to follow and remove some aws_mqtt5_listener_config settings from needing to be set during `aws_mqtt_protocol_adapter_new_from_5`.

bretambrose: We cannot destroy the listener until we know the client will no longer send us events.

source/request-response/protocol_adapter.c
```diff
@@ -0,0 +1,428 @@
...
+        .completion_user_data = subscribe_data,
+    };
+
+    if (aws_mqtt5_client_subscribe(adapter->client, &subscribe_view, &completion_options)) {
+        goto error;
+    }
+
+    return AWS_OP_SUCCESS;
+
+error:
```
sbSteveK, line 155: On subscribe, unsubscribe, and publish, the goto error results in the completion callback not being triggered. It's probably fine to let the Request Response client timeout the operations but we could also trigger the callback here and let them know right away that something went wrong. I forget off-hand how long the timeouts were but might be nice?

bretambrose: synchronous failures never invoke callbacks

source/request-response/protocol_adapter.c
sbSteveK: Let's keep it in if we can. This sort of description seems extremely useful for anyone coming to this fresh.

tests/request-response/request_response_protocol_adapter_tests.c
```diff
@@ -0,0 +1,1011 @@
...
+    aws_mqtt_library_init(allocator);
+
+    struct mqtt5_client_test_options test_options;
+    aws_mqtt5_client_test_init_default_options(&test_options);
+
+    if (test_type == PAOTT_SUCCESS) {
+        test_options.server_function_table.packet_handlers[AWS_MQTT5_PT_UNSUBSCRIBE] =
+            aws_mqtt5_mock_server_handle_unsubscribe_unsuback_success;
+    } else if (test_type == PAOTT_FAILURE_REASON_CODE) {
+        test_options.server_function_table.packet_handlers[AWS_MQTT5_PT_UNSUBSCRIBE] =
```
sbSteveK, line 573: Here and for subscribe and publish tests, the function should be set to a separate "timeout" function instead of the failed. On timeout, the client doesn't receive the same type of packet it would from the server and instead fails with `AWS_ERROR_MQTT_TIMEOUT` error. This would properly simulate what the adapter would be receiving and handling. If we don't do that, we probably don't need to have separate timeout failure tests because functionally they aren't taking any different path than failure error code and failure reason code.

include/aws/mqtt/private/request-response/protocol_adapter.h
```diff
@@ -0,0 +1,186 @@
...
+struct aws_protocol_adapter_publish_options {
+    struct aws_byte_cursor topic;
+    struct aws_byte_cursor payload;
+    uint32_t ack_timeout_seconds;
+
+    /*
+     * Invoked on success/failure of the publish itself.  Our implementations use QoS1 which means that success
+     * will be on puback receipt.
+     */
+    void (*completion_callback_fn)(bool, void *);
```
xiazhvera, line 55: It is probably useful to expose the error code for completion callback so that we can tell if it is a timeout or other client/server error. We can chose not to expose it to users in the service clients, but it would be nice to have at least have it in adapter so we could have more info.

tests/request-response/request_response_protocol_adapter_tests.c
bretambrose: I don't understand what you're saying here.  When it's a timeout test there's no packet handler so the operation times out.

include/aws/mqtt/private/request-response/protocol_adapter.h
bretambrose: I don't agree.  This is private/internal and the next level up does not need the reason for correct functionality; if it ever does, it can be added without issue.

tests/request-response/request_response_protocol_adapter_tests.c
sbSteveK: Ah, my mistake. I thought `s_aws_mqtt5_server_send_failed_unsuback_on_unsubscribe` was being assigned for all non successful PAOTT. The timeout tests are correctly checking against timeout.

include/aws/mqtt/private/request-response/protocol_adapter.h
bretambrose: Changed my mind.  It's not needed by the subscription manager but having the error code makes the tests more trustworthy because we can check why it failed.  I'll add it.

PR #346, opened 2024-01-26
Mqtt 311 Listener

bretambrose:
* Adds 311 client listener support - only incoming publish and connection success events are supported
* Reworks 311 protocol testing framework to be usable by different source files


source/mqtt311_listener.c
```diff
@@ -0,0 +1,291 @@
...
+    struct aws_mqtt311_callback_set_manager *manager,
+    struct aws_mqtt311_callback_set *callback_set) {
+
+    AWS_FATAL_ASSERT(
+        aws_event_loop_thread_is_callers_thread(s_mqtt_client_connection_get_event_loop(manager->connection)));
+
+    struct aws_mqtt311_callback_set_entry *entry = s_new_311_callback_set_entry(manager, callback_set);
+
+    aws_linked_list_push_front(&manager->callback_set_entries, &entry->node);
+
```
sbSteveK, line 210: Trivial: We have the created entry log above. Should we also have an "MQTT311 callback manager added entry id" log as well? Or should we just assume any entry created is immediately added?

source/mqtt311_listener.c
```diff
@@ -0,0 +1,291 @@
...
+    struct aws_mqtt311_callback_set_entry *entry =
+        aws_mem_calloc(manager->allocator, 1, sizeof(struct aws_mqtt311_callback_set_entry));
+
+    entry->allocator = manager->allocator;
+    entry->id = manager->next_callback_set_entry_id++;
+    entry->callbacks = *callback_set;
+
+    AWS_LOGF_INFO(
+        AWS_LS_MQTT_GENERAL,
+        "id=%p: MQTT311 callback manager created new entry :%" PRIu64,
```
sbSteveK, line 193: Trivial: Can we add "id" to line up with the log below? "created new entry with id=%"

source/mqtt311_listener.c
```diff
@@ -0,0 +1,291 @@
...
+void aws_mqtt311_callback_set_manager_on_publish_received(
+    struct aws_mqtt311_callback_set_manager *manager,
+    const struct aws_byte_cursor *topic,
+    const struct aws_byte_cursor *payload,
+    bool dup,
+    enum aws_mqtt_qos qos,
+    bool retain) {
+
+    struct aws_mqtt_client_connection_311_impl *connection_impl = manager->connection->impl;
+    AWS_FATAL_ASSERT(aws_event_loop_thread_is_callers_thread(connection_impl->loop));
```
sbSteveK, line 255: `s_mqtt_client_connection_get_event_loop` isn't used here and in the below function. I'm not positive it needs to be used here but I was also not 100% it needed to be used above so calling it out just in case it was just missed here.

sbSteveK (approved):

source/mqtt311_listener.c
bretambrose: There are no intervening points of failure so I don't think it would be useful.

source/mqtt311_listener.c
bretambrose: These functions need the downcasted impl anyways, so the loop can be pulled off easily after that.

PR #347, opened 2024-01-29
311 timeouts

bretambrose:
* Adds support and tests for 311-client-wide operation timeout for Subscribe and all of its variants.  The previous timeout implementation did not apply to subscribe.
* Adds internal APIs (and tests) for per-311-operation timeouts for Subscribe, Unsubscribe, and Publish


source/client.c
```diff
@@ -3025,6 +3111,21 @@ static uint16_t s_aws_mqtt_client_connection_311_publish(
...
+    const struct aws_byte_cursor *topic,
+    enum aws_mqtt_qos qos,
+    bool retain,
+    const struct aws_byte_cursor *payload,
+    aws_mqtt_op_complete_fn *on_complete,
+    void *userdata) {
+
+    struct aws_mqtt_client_connection_311_impl *connection = impl;
+
+    return aws_mqtt_client_connection_311_publish(
```
sbSteveK, line 3125: if `qos == 0` should we not set the `timeout_ns` to `connection->operation_timeout`? That may result in timeout tasks being generated for publishes that shouldn't have them. This might be caught elsewhere but I'm not immediately seeing it.

tests/v3/connection_state_test.c
```diff
@@ -2424,9 +2485,328 @@ static int s_test_mqtt_connection_unsub_timeout_fn(struct aws_allocator *allocat
...
+
+    struct aws_mqtt_connection_options connection_options = {
+        .user_data = state_test_data,
+        .clean_session = false,
+        .client_id = aws_byte_cursor_from_c_str("client1234"),
+        .host_name = aws_byte_cursor_from_c_str(state_test_data->endpoint.address),
+        .socket_options = &state_test_data->socket_options,
+        .on_connection_complete = aws_test311_on_connection_complete_fn,
+        .ping_timeout_ms = DEFAULT_TEST_PING_TIMEOUT_MS,
+        .protocol_operation_timeout_ms = 3000,
```
sbSteveK, line 2566: It looks like we're using the same value for the regular timeout tests and overridden timeout tests (3 seconds). Let's use different values (shorter for override) to make sure we're using the override value when examining test results. Also, might be useful to set the `protocol_operation_timeout_ms` in the override tests to make sure we're overriding the provided value explicitly. Not setting it defaults the regular value to `UINT64_MAX` which is ignored so any timeout occurring probably indicates this test is successful.

sbSteveK (approved):
Some trivial comments that may or may not need action but otherwise looks good.

source/client.c
```diff
@@ -1965,6 +1966,18 @@ static void s_subscribe_complete(
...
+
+    /*
+     * If we have a forward pointer to a timeout task, then that means the timeout task has not run yet.  So we should
+     * follow it and zero out the back pointer to us, because we're going away now.  The timeout task will run later
+     * and be harmless (even vs. future operations with the same packet id) because it only cancels if it has a back
+     * pointer.
+     */
+    if (task_arg->timeout_wrapper.timeout_task_arg) {
+        task_arg->timeout_wrapper.timeout_task_arg->task_arg_wrapper = NULL;
+        task_arg->timeout_wrapper.timeout_task_arg = NULL;
```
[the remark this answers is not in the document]
bretambrose: It does not appear to be useful.

source/client.c
bretambrose: There's an early out in the timeout scheduling function that ignores qos 0 publishes (because packet id is 0)

tests/v3/connection_state_test.c
bretambrose: I'm not bothered by this.  Elapsed time comparisons in tests have consisently blown up in our faces when containers and CPU throttling on cloud CI (and emulated ARM builds) get thrown into the mix.

PR #348, opened 2024-02-02
push ping time only if the request complete is successful

alfred2g:
push ping time only if the request complete is successful


tests/v3/connection_state_test.c
```diff
@@ -2997,6 +3000,7 @@ static int s_test_mqtt_connection_publish_QoS1_timeout_fn(struct aws_allocator *
         state_test_data);
     ASSERT_TRUE(packet_id_1 > 0);
 
+    uint64_t first_ping_time = connection->next_ping_time;
```
bretambrose, line 3003: It's not threadsafe to reach inside the connection if you're not in its event loop thread.

tests/v3/connection_state_test.c
```diff
@@ -3015,6 +3015,89 @@ AWS_TEST_CASE_FIXTURE(
...
+
+    /* Disable the auto ACK packets sent by the server, which blocks the requests to complete */
+    mqtt_mock_server_disable_auto_ack(state_test_data->mock_server);
+
+    /* make a publish with QoS 1 immediate. */
+    aws_mutex_lock(&state_test_data->lock);
+    state_test_data->expected_ops_completed = 1;
+    aws_mutex_unlock(&state_test_data->lock);
+
+    uint16_t packet_id_1 = aws_mqtt_client_connection_publish(
```
bretambrose, line 3051: My only concern is that it might be a little sensitive to runtime environment timing.  I would update it to something like this:

```
Repeat 10 times
   publish qos 1
   sleep 1 second

Verify at least one ping went out.
```

Auto ack setting does not affect pingresp generation, so it should be fine.

I don't think you need to verify any completion error codes on the publish (we disabled acks), only that they actually completed.

tests/v3/connection_state_test.c
```diff
@@ -3015,6 +3015,79 @@ AWS_TEST_CASE_FIXTURE(
...
+
+    ASSERT_SUCCESS(aws_mqtt_client_connection_connect(state_test_data->mqtt_connection, &connection_options));
+    s_wait_for_connection_to_complete(state_test_data);
+
+    /* Disable the auto ACK packets sent by the server, which blocks the requests to complete */
+    mqtt_mock_server_disable_auto_ack(state_test_data->mock_server);
+
+    /* make a publish with QoS 1 immediate. */
+    aws_mutex_lock(&state_test_data->lock);
+    state_test_data->expected_ops_completed = 1;
```
bretambrose, line 3048: bump to 10?

bretambrose (approved):
Fix n ship

PR #349, opened 2024-02-02
311 protocol adapter

bretambrose:
Impl and tests for 311 client protocol adapter.  Fewer tests than the 5 implementation because some packets don't have negative acks and there's no reasonable way to force an async failure into the 311 client.


source/request-response/protocol_adapter.c
```diff
@@ -31,66 +34,454 @@
...
 
-// used by both subscribe and unsubscribe
-struct aws_mqtt_protocol_adapter_5_subscription_op_data {
-    struct aws_allocator *allocator;
+/* Subscribe */
 
-    struct aws_byte_buf topic_filter;
-    struct aws_weak_ref *callback_ref;
+static void s_protocol_adapter_311_subscribe_completion(
+    struct aws_mqtt_client_connection *connection,
```
sbSteveK, line 120: do connection, topic, and packet_id need to be part of this function call? It doesn't appear like they're used in any context and could be removed and added if they're ever needed in the future.

sbSteveK: Scratch that, we're passing the function to the 311 subscribe callback -_-;;

PR #351, opened 2024-02-28
Request Response Subscription manager

bretambrose:


source/mqtt.c
```diff
@@ -254,6 +254,7 @@ static struct aws_error_info_list s_error_list = {
             DEFINE_LOG_SUBJECT_INFO(AWS_LS_MQTT5_CLIENT, "mqtt5-client", "MQTT5 client and connections"),
             DEFINE_LOG_SUBJECT_INFO(AWS_LS_MQTT5_CANARY, "mqtt5-canary", "MQTT5 canary logging"),
             DEFINE_LOG_SUBJECT_INFO(AWS_LS_MQTT5_TO_MQTT3_ADAPTER, "mqtt5-to-mqtt3-adapter", "MQTT5-To-MQTT3 adapter logging"),
+            DEFINE_LOG_SUBJECT_INFO(AWS_LS_MQTT_REQUEST_RESPONSE, "mqtt-request-response-client", "MQTT request-response client logging"),
```
sbSteveK, line 257: Trivial: Should this be "AWS_LS_MQTT_REQUEST_RESPONSE_CLIENT" ?

source/request-response/subscription_manager.c
```diff
@@ -0,0 +1,624 @@
...
+
+static void s_apply_session_lost(struct aws_rr_subscription_manager *manager) {
+    aws_hash_table_foreach(&manager->subscriptions, s_apply_session_lost_wrapper, manager);
+}
+
+void aws_rr_subscription_manager_on_protocol_adapter_connection_event(
+    struct aws_rr_subscription_manager *manager,
+    const struct aws_protocol_adapter_connection_event *event) {
+
+    AWS_LOGF_DEBUG(
```
sbSteveK, line 572: Trivial: We should probably move this log into the if statements below `if (event->event_type == AWS_PACET_CONNECTED)` and exclude the "joined_session %d" for disconnected connection events for slightly cleaner logging. Or at least explicitly set the `event->joined_session` to false when creating the aws_protocol_adapter_connection_event struct before it is passed here rather than potentially taking on whatever that value was on a non-zeroed struct.

include/aws/mqtt/private/request-response/subscription_manager.h
```diff
@@ -0,0 +1,206 @@
...
+ */
+struct aws_rr_subscription_manager {
+    struct aws_allocator *allocator;
+
+    struct aws_rr_subscription_manager_options config;
+
+    /* non-owning reference; the client is responsible for destroying this asynchronously (listener detachment) */
+    struct aws_mqtt_protocol_adapter *protocol_adapter;
+
+    /* &aws_rr_subscription_record.topic_filter_cursor -> aws_rr_subscription_record * */
```
sbSteveK, line 88: Needs cleanup?

bretambrose: Not following; that's a type definition doc comment

source/mqtt.c
bretambrose: Think Id rather drop the client part actually.

bretambrose: Actually would prefer to just get rid of client entirely

include/aws/mqtt/private/request-response/subscription_manager.h
bretambrose: It's a type annotation.

include/aws/mqtt/private/request-response/subscription_manager.h
```diff
@@ -0,0 +1,206 @@
...
+    struct aws_hash_table subscriptions;
+
+    bool is_protocol_client_connected;
+};
+
+enum aws_rr_subscription_type {
+    ARRST_EVENT_STREAM,
+    ARRST_REQUEST_RESPONSE,
```
xiazhvera, line 96: Trivial: I realized the naming `ARRST_REQUEST_RESPONSE` is confusing as the client is called requrest-respond client, while it has a `request-respond`operation and a `steaming` operation. Probably use names like : single request-respond v.s. streaming, RESTful v.s. Streaming?

source/request-response/subscription_manager.c
```diff
@@ -0,0 +1,627 @@
...
+        AWS_LS_MQTT_REQUEST_RESPONSE,
+        "request-response subscription manager - unsubscribe submitted for ('" PRInSTR "')",
+        AWS_BYTE_CURSOR_PRI(record->topic_filter_cursor));
+
+    record->pending_action = ARRSPAT_UNSUBSCRIBING;
+}
+
+/* Only called when shutting down the client */
```
sbSteveK, line 165: Trivial: client->manager

bretambrose: Actually that's intentional.  The (subscription manager) function is only called when the whole request-response client is getting shut down.

include/aws/mqtt/private/request-response/subscription_manager.h
```diff
@@ -0,0 +1,206 @@
...
+
+    /*
+     * Maximum number of concurrent subscriptions allowed
+     */
+    size_t max_subscriptions;
+
+    /*
+     * Ack timeout to use for all subscribe and unsubscribe operations
+     */
+    uint32_t operation_timeout_seconds;
```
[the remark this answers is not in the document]
bretambrose: No, at present there's a single timeout for all operations.  We added per-operation because we can't control if the mqtt client even has a timeout enabled, let alone what it is.

include/aws/mqtt/private/request-response/subscription_manager.h
bretambrose: It's mildly confusing but I haven't seen a better name.

source/request-response/subscription_manager.c
```diff
@@ -0,0 +1,627 @@
...
+}
+
+enum aws_acquire_subscription_result_type aws_rr_subscription_manager_acquire_subscription(
+    struct aws_rr_subscription_manager *manager,
+    const struct aws_rr_acquire_subscription_options *options) {
+    struct aws_rr_subscription_record *existing_record = s_get_subscription_record(manager, options->topic_filter);
+
+    // is no subscription present?
+    if (existing_record == NULL) {
+        s_cull_unused_subscriptions(manager);
```
sbSteveK, line 386: I forgot where we landed on this and can't easily find it in the tech spec but are we going to run the cull every time a new subscription record is requested? Or should we keep unused request-response subs until we need the space?

bretambrose: We're doing lazy culling on subscribe regardless of what the tech doc says.

sbSteveK:
Just a random thought but would there be any value in adding an 'unsubscribe to all' function that can be called prior to destruction? The unsubscribes would be sent to the mqtt client to be processed (or not processed if the client is also being destroyed) to free up whatever subscriptions the RR had. Would be fire and forget with no wait on acks. This may be useful in a use-case where someone is creating and destroying multiple rr clients over the course of a connection with different names.

tests/CMakeLists.txt
```diff
@@ -481,6 +481,32 @@ add_test_case(mqtt311_listener_connection_events_no_session)
 add_test_case(mqtt311_listener_connection_events_with_session)
 add_test_case(mqtt311_listener_publish_event)
 
+# "rrsm" = "request_response_subscription_manager"
```
sbSteveK, line 484: Trivial: The `aws_rr_subscription_manager_init()` would immediately return an `AWS_ERROR_INVALID_ARGUMENT` error but should we still add a test that insures a max_subscriptions < 1 fails?

bretambrose:
> Just a random thought but would there be any value in adding an 'unsubscribe to all' function that can be called prior to destruction? The unsubscribes would be sent to the mqtt client to be processed (or not processed if the client is also being destroyed) to free up whatever subscriptions the RR had. Would be fire and forget with no wait on acks. This may be useful in a use-case where someone is creating and destroying multiple rr clients over the course of a connection with different names.

That's what `aws_rr_subscription_manager_clean_up` does

sbSteveK (approved):

source/request-response/subscription_manager.c
bretambrose: As a follow-up, this is not quite accurate anymore.  Subscriptions are purged eagerly and do not linger indefinitely.

PR #355, opened 2024-03-04
Request response client - create/destroy/lifecycle

bretambrose:


source/request-response/request_response_client.c
```diff
@@ -0,0 +1,388 @@
...
+    struct aws_rr_subscription_manager_options sm_options = {
+        .max_subscriptions = options->max_subscriptions,
+        .operation_timeout_seconds = options->operation_timeout_seconds,
+    };
+
+    /*
+     * We can't initialize the subscription manager until we're running on the event loop, so make sure that
+     * initialize can't fail by checking its options for validity now.
+     */
+    if (!aws_rr_subscription_manager_are_options_valid(&sm_options)) {
```
sbSteveK, line 220: Note: Maybe it's already in place somewhere or we simply haven't gotten around to it yet but just noting that an `operation_timeout_seconds` of 0 results in this failing. Wondering if there's a point before this where a default is applied to that value if it hasn't been set or if that default should be applied during the validity check.

source/request-response/request_response_client.c
```diff
@@ -0,0 +1,388 @@
...
+
+    struct aws_mqtt_protocol_adapter_options adapter_options = {
+        .subscription_event_callback = s_aws_rr_client_protocol_adapter_subscription_event_callback,
+        .incoming_publish_callback = s_aws_rr_client_protocol_adapter_incoming_publish_callback,
+        .terminate_callback = s_aws_rr_client_protocol_adapter_terminate_callback,
+        .connection_event_callback = s_aws_rr_client_protocol_adapter_connection_event_callback,
+        .user_data = rr_client,
+    };
+
+    rr_client->client_adapter = aws_mqtt_protocol_adapter_new_from_311(rr_client->allocator, &adapter_options, client);
```
sbSteveK, line 322: `s_aws_mqtt_request_response_client_new` can return null if `sm_options` fails validation. We need a null check on the `rr_client` before assigning a `client_adapter`

source/request-response/request_response_client.c
```diff
@@ -0,0 +1,388 @@
...
+    return NULL;
+}
+
+struct aws_mqtt_request_response_client *aws_mqtt_request_response_client_new_from_mqtt5_client(
+    struct aws_allocator *allocator,
+    struct aws_mqtt5_client *client,
+    const struct aws_mqtt_request_response_client_options *options) {
+
+    struct aws_mqtt_request_response_client *rr_client =
+        s_aws_mqtt_request_response_client_new(allocator, options, client->loop);
```
sbSteveK, line 345: Same as above. Need a null check after this.

tests/request-response/request_response_client_tests.c
```diff
@@ -0,0 +1,233 @@
...
+
+enum rr_test_client_protocol {
+    RRCP_MQTT311,
+    RRCP_MQTT5,
+};
+
+struct aws_rr_client_test_fixture {
+    struct aws_allocator *allocator;
+
+    struct aws_mqtt_request_response_client *client;
```
sbSteveK, line 22: Trivial: For a little added clarity in tests, can we rename this to `rr_client` or something that explicitly separates this from the mqtt_client that is also being referenced in tests?

sbSteveK (approved):
Looks solid outside of potentially missing null checks.

source/request-response/request_response_client.c
bretambrose: Timeout support is next.  Based on our discussion, I'm thinking about the following:

* Split operation_timeout_seconds into (request_response_timeout_seconds and streaming_subscribe_timeout_seconds)
* A zero-valued request response timeout gets turned into a default (30 or 60)
* A zero-value streaming timeout means no timeout

Does the inconsistency between how zero is treated by each setting outweigh the fact that we really want to have a reasonable default for request-response timeout and no timeout default for streaming operations?

PR #357, opened 2024-03-13
Refactor to support new durable streaming operations

bretambrose:
Update subscription manager and associated systems to support durable/indefinite streaming operations


source/request-response/subscription_manager.c
```diff
@@ -297,26 +307,55 @@ static void s_cull_unused_subscriptions(struct aws_rr_subscription_manager *mana
     aws_hash_table_foreach(&manager->subscriptions, s_rr_subscription_cull_unused_subscriptions_wrapper, manager);
 }
 
-static const char *s_request_response_subscription_event_type_to_c_str(enum aws_rr_subscription_event_type type) {
+static const char *s_rr_subscription_event_type_to_c_str(enum aws_rr_subscription_event_type type) {
```
sbSteveK, line 310: Trivial: In other recent cases where we're returning a readable string, it feels like we don't translate the enum to a prefix-less PascalCase but instead return a normal reading format. e.g. "Request subscribe successful" vs. "RequestSubscribeSuccess". I think that's preferable but it's extremely trivial and doesn't need to be changed but might be better for logs.

source/request-response/subscription_manager.c
```diff
@@ -344,6 +383,10 @@ static int s_rr_activate_idle_subscription(
     struct aws_rr_subscription_record *record) {
     int result = AWS_OP_SUCCESS;
 
+    if (record->poisoned) {
+        return AWS_OP_SUCCESS;
```
sbSteveK, line 387: Should we log that this is a poisoned success? It may help track down poison related issues in the future.

sbSteveK (approved):

source/request-response/subscription_manager.c
bretambrose: Have come to prefer camel case for easier searchability, tbh.

source/request-response/subscription_manager.c
bretambrose: Yeah that's a good idea.

PR #359, opened 2024-03-14
Update README to accurately reflect what is copied

Octogonapus:
The README was not updated when changes were made to copy parameters such as the topic filter and payload for various functions.
This PR fixes those instances so that the documentation is accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ int aws_mqtt_client_connection_set_will(
 ```
 Sets the last will and testament to be distributed by the server upon client disconnection. Must be called before
 `aws_mqtt_client_connection_connect`. See `aws_mqtt_client_connection_publish` for information on the parameters.
-`topic` and `payload` must persist past the call to `aws_mqtt_client_connection_connect`.
+`topic` and `payload` are copied.
 
 ```c
 int aws_mqtt_client_connection_set_login(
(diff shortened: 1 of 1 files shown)


bretambrose:
Honestly, most of that README should be removed.  We don't want people using the 311 implementation and this is not a stable API in general.

Octogonapus:
Why don't you want people using the v3.1.1 implementation? The AWS broker still supports v3.1.1. Is it being phased out in favor of the v5 to v3.1.1 adapter?

bretambrose:
> Why don't you want people using the v3.1.1 implementation? The AWS broker still supports v3.1.1. Is it being phased out in favor of the v5 to v3.1.1 adapter?

Nothing's being phased out at the moment.

The 311 implementation was the first client that got written on top of the CRT event loop APIs.  As such it has a lot of properties/decisions that we've come to regret.
* The public API exposes things that it has no business doing (synchronous packet id allocation, for example).  Other properties (per-subscription callbacks, for example) are nice features in theory, but they do not belong in the low-level protocol client.
* The implementation makes heavy usage of shared state and the locking around that state is a nightmare to reason about.  I dread anything that forces me to make changes in the 311 implementation.
* The delicate shared state usage makes it much harder to add new features.

In my ideal world, everyone would use the 5 client.  It's not perfect (as I work on another rust MQTT client as a personal project, I keep finding small things that could be improved) but it is much more maintainable and easy to reason about.  Maybe someday we can extend the current 5 client to also have a 311 mode while keeping the current API contract and sunset the 311 implementation, but that's just a pipe dream atm.

Octogonapus:
Thank you for explaining that!

PR #361, opened 2024-03-22
Refactor MQTT5 ping timeout

bretambrose:
* Relaxes validation on ping timeout configuration value.
* Ping timeout now a function of base configuration value and the connection's negotiated keep alive.


source/v5/mqtt5_client.c
```diff
@@ -2940,7 +2940,20 @@ static void s_on_pingreq_send(struct aws_mqtt5_client *client) {
...
+        aws_timestamp_convert(
+            client->negotiated_settings.server_keep_alive, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL) /
+        2;
+
+    uint64_t connection_ping_timeout = ping_timeout_nanos;
+    if (connection_ping_timeout == 0 || connection_ping_timeout > half_keep_alive_nanos) {
+        connection_ping_timeout = half_keep_alive_nanos;
+    }
+
+    AWS_LOGF_DEBUG(
```
sbSteveK, line 2953: Trivial: Should this log only occur when the set ping timout is dynamically altered? i.e. in the if block that checks whether we should apply the `half_keep_alive_nanos` ?

sbSteveK (approved):

source/v5/mqtt5_client.c
bretambrose: I think it's fine the way it is as a direct reminder so that you don't have to have access to the log of the original settings to know.

> Trivial: Should this log only occur when the set ping timout is dynamically altered? i.e. in the if block that checks whether we should apply the `half_keep_alive_nanos` ?

PR #362, opened 2024-03-26
Streaming operations

bretambrose:
* Reworks subscription manager event set to match what's needed by the request-response client
* Request-response streaming operation support and tests


source/request-response/request_response_client.c
```diff
@@ -564,9 +714,129 @@ static void s_mqtt_request_response_client_wake_service(struct aws_mqtt_request_
...
+    struct aws_task task;
+
+    struct aws_mqtt_request_response_client *rr_client;
+
+    enum aws_rr_subscription_event_type type;
+    struct aws_byte_buf topic_filter;
+    uint64_t operation_id;
+};
+
+static void s_aws_rr_subscription_status_event_task_delete(struct aws_rr_subscription_status_event_task *task) {
```
sbSteveK, line 735: Trivial: Why delete instead of destroy?

source/request-response/request_response_client.c
```diff
@@ -796,6 +1130,145 @@ static uint64_t s_mqtt_request_response_client_get_next_service_time(struct aws_
...
+    struct aws_mqtt_rr_client_operation *operation) {
+    if (operation->type == AWS_MRROT_REQUEST) {
+        return operation->storage.request_storage.options.subscription_topic_filter;
+    } else {
+        return operation->storage.streaming_storage.options.topic_filter;
+    }
+}
+
+/* TODO: add aws-c-common API? */
+static bool s_is_operation_in_list(const struct aws_mqtt_rr_client_operation *operation) {
```
sbSteveK, line 1143: Trivial: If this is added to aws-c-common it'll have a description that clarifies what it's doing but just looking at the name of the function it's not immediately clear what it's doing. Or maybe it is since its only argument is an operation and not a list to check whether the operation is in that list or not. Either way, never thought about a check for whether a node is part of a list or hasn't been added to one yet.

source/request-response/request_response_client.c
```diff
@@ -796,6 +1130,145 @@ static uint64_t s_mqtt_request_response_client_get_next_service_time(struct aws_
...
+            (void *)client,
+            AWS_BYTE_CURSOR_PRI(topic_filter_cursor));
+    } else {
+        entry = element->value;
+    }
+
+    AWS_FATAL_ASSERT(entry != NULL);
+
+    if (s_is_operation_in_list(operation)) {
+        aws_linked_list_remove(&operation->node);
```
sbSteveK, line 1174: It's unfortunate that `aws_linked_list_remove` doesn't do the check you implemented in `s_is_operation_in_list` as a prerequisite to linking the prev and next nodes together. It's weird that instead it crashes in debug but just runs in release.

source/request-response/request_response_client.c
```diff
@@ -1259,6 +1731,78 @@ void s_aws_mqtt_request_operation_storage_init_from_options(
...
+    struct aws_mqtt_rr_client_operation *operation,
+    struct aws_mqtt_request_response_client *client) {
+    struct aws_logger *log_handle = aws_logger_get_conditional(AWS_LS_MQTT_REQUEST_RESPONSE, AWS_LL_DEBUG);
+    if (log_handle == NULL) {
+        return;
+    }
+
+    struct aws_mqtt_request_operation_options *options = &operation->storage.request_storage.options;
+
+    AWS_LOGUF(
```
sbSteveK, line 1744: Trivial: I've seen grouped logs like this elsewhere and just wondering if it would make sense to combine all these individual logs into a single log instead of a stream of a half dozen that may be broken up by other logs. No biggie since if we combined them, they'd probably take up the same amount of vertical space in the logs anyway.

source/request-response/subscription_manager.c
```diff
@@ -326,6 +343,12 @@ static const char *s_rr_subscription_event_type_to_c_str(enum aws_rr_subscriptio
 
         case ARRSET_STREAMING_SUBSCRIPTION_HALTED:
             return "StreamingSubscriptionHalted";
+
+        case ARRSET_UNSUBSCRIBE_COMPLETE:
+            return "UnsubscribeComplete";
```
sbSteveK, line 348: Trivial: the `s_aws_mqtt_request_response_operation_state_to_c_str` returns in the form of "PENDING_RESPONSE" and here we're returning in the form of pascal case... nvm, while writing this I found that we're using snake case elsewhere and there's no conformity =P

sbSteveK (approved):

source/request-response/request_response_client.c
bretambrose: Well it would be is_node_in_list which feels somewhat self-explanatory

bretambrose: https://github.com/awslabs/aws-c-common/pull/1103

PR #365, opened 2024-04-12
Support multiple subscriptions in request operations

bretambrose:
* Refactor request operations to support multiple subscription filters


source/request-response/subscription_manager.c
```diff
@@ -451,114 +443,150 @@ static int s_rr_activate_idle_subscription(
...
-            AWS_BYTE_CURSOR_PRI(options->topic_filter),
-            options->operation_id);
-        return AASRT_FAILURE;
-    }
+    for (size_t i = 0; i < options->topic_filter_count; ++i) {
+        struct aws_byte_cursor topic_filter = options->topic_filters[i];
+        struct aws_rr_subscription_record *existing_record = s_get_subscription_record(manager, topic_filter);
 
-    s_aws_rr_subscription_record_log_invariant_violations(existing_record);
+        if (s_rr_activate_idle_subscription(manager, existing_record)) {
```
sbSteveK, line 576: If this is failing on the first of a number of topics, does no further action need to be taken on subsequent topics? e.g. we emit a subscription event for the failing record. Should we also emit the subscription event for remaining records or notify that x number of subscription attempts are being skipped due to this one's failure?

sbSteveK (approved):

source/request-response/subscription_manager.c
bretambrose: No, one failure causes the operation to fail which in turn causes the operation to release all its acquisition attempts

source/request-response/subscription_manager.c
```diff
@@ -748,7 +775,7 @@ void aws_rr_subscription_manager_on_protocol_adapter_connection_event(
 }
 
 bool aws_rr_subscription_manager_are_options_valid(const struct aws_rr_subscription_manager_options *options) {
-    if (options == NULL || options->max_subscriptions < 1 || options->operation_timeout_seconds == 0) {
+    if (options == NULL || options->max_request_response_subscriptions < 2 || options->operation_timeout_seconds == 0) {
```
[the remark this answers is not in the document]
bretambrose: I don't see a point in supporting that.  If you don't use request response then that topic space is unused.

PR #367, opened 2024-04-16
Final request-response polish

bretambrose:
Full MQTT request-response client and subsystem implementation


source/request-response/subscription_manager.c
```diff
@@ -445,7 +445,12 @@ enum aws_acquire_subscription_result_type aws_rr_subscription_manager_acquire_su
     const struct aws_rr_acquire_subscription_options *options) {
 
     if (options->topic_filter_count == 0) {
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        AWS_LOGF_ERROR(
```
sbSteveK, line 448: Don't feel especially strong on this but the result of `aws_rr_subscription_manager_acquire_subscription` are not being treated as errors and instead just a failed attempt to acquire a subscription for various reasons. If we aren't throwing an error for the issues of a poisoned or type missmatch, we should log these at a Debug level rather than an Error.

sbSteveK (approved):

source/request-response/subscription_manager.c
bretambrose: Added explanatory comment

PR #369, opened 2024-05-15
Raise error on null mqtt5 client

xiazhvera:
*Issue #, if available:*

*Description of changes:*


sfod:
This PR relaxes the requirements for mqtt5 client to be valid, for *some* operations. Is it possible that this'll cause something bad in operations that don't validate client (e.g. https://github.com/awslabs/aws-c-mqtt/blob/v0.10.4/source/v5/mqtt5_client.c#L2109)?

xiazhvera:
The PR was to workaround the swift error issue. As there are some concern about adding the error in c-mqtt, I will re-evaluate the solution. Temporary close the ticket.

PR #381, opened 2025-01-30
Support wildcards streams in request-response stream client

sfod:
*Issue #, if available:*

Request-response stream needs support subscriptions to topics like `$aws/some-service/<EXECUTION_ID>/etc`, where `<EXECUTION_ID>` changes for each message published by a service.

*Description of changes:*

Implement a module allowing to match topic filter with wildcards and a topic where PUBLISH message was received.


source/request-response/request_response_subscription_set.c
```diff
@@ -0,0 +1,326 @@
...
+            }
+
+            if (aws_byte_cursor_eq_c_str(&subscription_topic_filter_segment, "#")) {
+                multi_level_wildcard = true;
+                match = true;
+                break;
+            }
+
+            if (!aws_byte_cursor_eq_c_str(&subscription_topic_filter_segment, "+") &&
+                !aws_byte_cursor_eq_ignore_case(&topic_segment, &subscription_topic_filter_segment)) {
```
bretambrose, line 267: I don't think case insensitive is right here

source/request-response/request_response_subscription_set.c
```diff
@@ -0,0 +1,327 @@
...
+
+    return entry;
+}
+
+struct aws_rr_operation_list_topic_filter_entry *aws_mqtt_request_response_client_subscriptions_add_stream_subscription(
+    struct aws_request_response_subscriptions *subscriptions,
+    const struct aws_byte_cursor *topic_filter) {
+    AWS_FATAL_ASSERT(subscriptions);
+
+    bool is_topic_with_wildcard =
```
bretambrose, line 143: If we don't have a helper for this, maybe we should make one.

source/request-response/request_response_subscription_set.c
```diff
@@ -0,0 +1,327 @@
...
+        struct aws_byte_cursor subscription_topic_filter_segment;
+        AWS_ZERO_STRUCT(subscription_topic_filter_segment);
+
+        struct aws_byte_cursor topic_segment;
+        AWS_ZERO_STRUCT(topic_segment);
+
+        bool match = true;
+        bool multi_level_wildcard = false;
+
+        while (aws_byte_cursor_next_split(&entry->topic_filter_cursor, '/', &subscription_topic_filter_segment)) {
```
bretambrose, line 255: Can we split the match out into a private helper function?

bretambrose (approved):
Couple minor requests

source/request-response/request_response_subscription_set.c
sfod: I was thinking about adding a new public function to the `mqtt.h` header, near `aws_mqtt_is_valid_topic` for example. But in the end I felt it addresses a local problem, so I've added a local helper function.

PR #383, opened 2025-02-19
Fix handling of websocket failed transform for MQTT311 connection

sfod:
*Issue #, if available:*
#V1680580336

When websocket transform fails, shutdown of the MQTT311 connection might happen on a thread different from this connection's event loop. In older releases it was allowed, however in https://github.com/awslabs/aws-c-mqtt/pull/367 we added a new restriction that shutdown of MQTT311 connection must happen on connection's event loop. This leads to `AWS_FATAL_ASSERT` statement terminating a process.

*Description of changes:*

MQTT311 connection shutdown always happens on connection's event loop.

### DIFF
--- a/source/client.c
+++ b/source/client.c
@@ -1364,6 +1364,7 @@ error:
 }
 
 struct mqtt_on_websocket_setup_task_arg {
+    struct aws_allocator *allocator;
     struct aws_task task;
     struct aws_mqtt_client_connection_311_impl *connection;
     int error_code;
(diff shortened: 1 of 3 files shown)


tests/v3/connection_state_test.c
```diff
@@ -3968,3 +3968,40 @@ AWS_TEST_CASE_FIXTURE(
...
+
+static void s_mqtt_client_test_websocket_failed_transform(
+    struct aws_http_message *request,
+    void *user_data,
+    aws_mqtt_transform_websocket_handshake_complete_fn *complete_fn,
+    void *complete_ctx) {
+
+    (void)user_data;
+
+    (*complete_fn)(request, AWS_ERROR_INVALID_STATE, complete_ctx);
```
bretambrose, line 3980: Is this happening outside the connection's event loop?

sfod: Yes, this is happening on the main thread. I can fire up a new thread that'll call `complete_fn`, so it's transparent the callback is not executed on the event loop. Or maybe just add a comment?

source/client.c
```diff
@@ -1363,6 +1363,47 @@ static int s_websocket_connect(struct aws_mqtt_client_connection_311_impl *conne
...
+    AWS_FATAL_ASSERT(connection);
+
+    /* s_on_websocket_setup shutdowns mqtt connection, this must happen on the MQTT connection's event loop. */
+    if (aws_event_loop_thread_is_callers_thread(connection->loop)) {
+        struct aws_websocket_on_connection_setup_data websocket_setup = {.error_code = error_code};
+        s_on_websocket_setup(&websocket_setup, connection);
+    } else {
+        struct mqtt_on_websocket_setup_task_arg *on_websocket_setup_task_arg =
+            aws_mem_calloc(connection->allocator, 1, sizeof(struct mqtt_on_websocket_setup_task_arg));
+        on_websocket_setup_task_arg->connection = connection;
```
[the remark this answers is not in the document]
sfod: I've added a comment providing a short description when connection instance is acquired and when released

PR #389, opened 2025-04-04
Capture fields from PUBLISH for req-resp stream operations

sfod:
*Issue #, if available:*

It's the same PR as https://github.com/awslabs/aws-c-mqtt/pull/386
That PR was reverted to unblock release for aws-c-mqtt. This time this PR will be merged only when all downstream projects are ready.

Certain service clients need additional data from MQTT PUBLISH packet.
Currently, the following fields should be added:
- content type
- user properties
- message expiry interval seconds

In future, more data may be added.

Description of changes:
Pass all these fields to callback as a struct.


include/aws/mqtt/request-response/request_response_client.h
```diff
@@ -31,6 +31,22 @@ struct aws_mqtt_request_operation_response_path {
     struct aws_byte_cursor correlation_token_json_path;
 };
 
+/*
+ * An event emitted by a streaming operation's subscription.
+ */
+struct aws_mqtt_request_response_publish_event {
```
xiazhvera, line 37: Trivial, I feel it makes more sense to name it `aws_mqtt_rr_incoming_publish_event`.

bretambrose (approved):

include/aws/mqtt/request-response/request_response_client.h
sfod: Similar to `aws_mqtt_rr_client_operation`? The name indeed doesn't follow any existing naming pattern:
- the streaming-related entity names start with `aws_mqtt_streaming_operation_`
- the request-related entity name start with `aws_mqtt_request_operation_`
- the entities related to req-resp client start with `aws_mqtt_request_response_client_` and `aws_mqtt_rr_client_`

Though, considering that this struct is used for streaming operations, maybe it should be named `aws_mqtt_streaming_operation_incoming_publish_event`? We even have `aws_mqtt_streaming_operation_incoming_publish_fn` function that gets this struct.

But this struct is also used in the internals of req-resp for both streaming and req-resp operations. Though, these are internals and don't matter mach in the naming sense.

xiazhvera: As the struct is used for both streaming operation and request response, therefore using prefix "aws_mqtt_rr_client_" makes more sense to me. Regardless, like you mentioned it is doesn't matter much, feel free to use your own judgment here!

sfod: OK, I've added this struct to request operations. And now since it's used for both streaming and request operations, the `aws_mqtt_rr_incoming_publish_event` name is perfect. Thanks for bringing this up 👍

PR #395, opened 2025-05-06
Update test to check time lower bound

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/v5/mqtt5_client_tests.c
+++ b/tests/v5/mqtt5_client_tests.c
@@ -18,10 +18,20 @@
 #include <aws/testing/aws_test_harness.h>
 
 #include <math.h>
+#define AWS_MQTT5_TESTING_DELAY_NS 200 * 1000 * 1000 /*200ms*/
 
 static bool s_is_within_percentage_of(uint64_t expected_time, uint64_t actual_time, double percentage) {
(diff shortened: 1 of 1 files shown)


tests/v5/mqtt5_client_tests.c
```diff
@@ -1717,7 +1727,7 @@ static int s_verify_reconnection_after_success_used_backoff(
         AWS_TIMESTAMP_MILLIS,
         NULL);
 
-    if (!s_is_within_percentage_of(expected_reconnect_delay_ms, post_success_reconnect_time_ms, .3)) {
+    if (!s_validate_time_interval(expected_reconnect_delay_ms, post_success_reconnect_time_ms, .3)) {
```
sfod, line 1730: Shouldn't we use here the lower bound check as well? Currently it seems that it's a valid case when the last reconnect happens slightly earlier than anticipated, and it's ok. I can't figure out why.

sfod:
I'd add a few short comments explaining why these fixes are here, for future us.

tests/v5/mqtt5_client_tests.c
sfod: Aha, with a new comment I see that the upper bound check is still needed here

sfod (approved):

PR #400, opened 2025-06-12
Request Response Duplicate Correlation Token Validation

sbSteveK:
If a Request-Response Operation attempts to use an already in-use correlation token it will now fail validation on submission of the operation to the RR Client.

Add a test for failing validation when a duplicate correlation token is submitted while the token is already in use.
Add a test for successful validation of a reused correlation token if the previous operation with the same correlation token has already completed.


bretambrose (approved):

source/request-response/request_response_client.c
```diff
@@ -1995,6 +2045,15 @@ int aws_mqtt_request_response_client_submit_request(
         &operation->storage.request_storage, allocator, request_options);
     s_aws_mqtt_rr_client_operation_init_shared(operation, client);
 
+    // Store the correlation token to prevent allowing a duplicate correlation token to be used.
+    if (operation->storage.request_storage.options.correlation_token.len > 0) {
+        aws_hash_table_put(
+            &client->correlation_tokens_in_use,
```
sfod: `aws_hash_table_put` and `aws_hash_table_remove` seem to be executed from different threads (a user thread and an event loop thread correspondingly), i.e. there is a possible race condition.

source/request-response/request_response_client.c
```diff
@@ -394,6 +419,8 @@ static void s_complete_request_operation_with_failure(struct aws_mqtt_rr_client_
 
     s_change_operation_state(operation, AWS_MRROS_PENDING_DESTROY);
 
+    s_remove_correlation_token_from_in_use(operation);
```
sfod: nit: Should we call `s_remove_correlation_token_from_in_use` only when token is non-empty? It's not an error from what I see, but will be more consistent.

source/request-response/request_response_client.c
sbSteveK: Moved addition of correlation token to `s_mqtt_rr_client_submit_operation` which is the function executed when the scheduled operation is executed. 

There is another check done here before adding the token because it's possible that two operations are scheduled using the same token but won't be detected till this point. This required the addition of a new error code for duplicate correlation tokens as we have to fail an operation at this point with an error and it should make it much clearer what went wrong.

source/request-response/request_response_client.c
sbSteveK: Unnecessary but I don't see why not. I've wrapped it in a len check on the token.

source/request-response/request_response_client.c
```diff
@@ -1616,6 +1657,17 @@ static bool s_are_request_operation_options_valid(
...
+        // do a correlation token check
+        if (s_is_correlation_token_in_use(client, &request_options->correlation_token)) {
+            AWS_LOGF_ERROR(
+                AWS_LS_MQTT_REQUEST_RESPONSE,
+                "(%p) rr client request options - correlation token is already in use.",
+                (void *)client);
+            return false;
+        }
+    }
+
```
sfod, line 1670: I think this block should be removed because `s_are_request_operation_options_valid` is executed on a user thread

source/request-response/request_response_client.c
```diff
@@ -856,6 +886,8 @@ static void s_complete_operation_with_correlation_token(
     struct aws_mqtt_rr_client_operation *operation = hash_element->value;
     AWS_FATAL_ASSERT(operation->type == AWS_MRROT_REQUEST);
 
+    s_remove_correlation_token_from_in_use(operation);
+
     if (operation->state == AWS_MRROS_PENDING_DESTROY) {
```
sfod, line 891: I think `s_remove_correlation_token_from_in_use` should be called only once in `s_remove_operation_from_client_tables`. I.e. an operation's token should become available for reuse only when there is no possibility for the operation to do something with it.

With current implementation, it's hard to reason about if it's possible or not for the previous operation to do something with its token when the token is already assigned to the next operation.

For example, take the following scenario.

Imagine there is an existing operation `op_1` with a token `token`.

- a new operation `op_2` with a token `token` started by a user thread, so `aws_mqtt_request_response_client_submit_request` executes:
  - schedule a task (for the next event loop iteration) which will perform this new operation initialization

event loop iteration N:
- a PUBLISH arrives for `op_1`, so `s_complete_operation_with_correlation_token` executes:
  - rm `token` from `correlation_tokens_in_use`
  - ... perform the rest of the function
- Initialize `op_2`, i.e. execute `s_mqtt_rr_client_submit_operation` function:
  - add `token` to `correlation_tokens_in_use`
  - ... and do other stuff

event loop iteration N + 1:
- the second PUBLISH arrives for `op_1` (or alternatively there was an op failure on the previous iteration)
  - rm `token` from `correlation_tokens_in_use`

Now there is `op_2` with a token `token`, but no `token` record in `correlation_tokens_in_use`.

To address this hypothetical scenario, we could just execute `s_remove_correlation_token_from_in_use(operation);` only if state not `AWS_MRROS_PENDING_DESTROY`. But I feel that this will be still kinda fragile.

source/request-response/request_response_client.c
sbSteveK: Removing!

source/request-response/request_response_client.c
sbSteveK: makes sense, changing now.

sbSteveK: I think the reason I went with the way I did was that we need to have a separate check in `s_complete_request_operation_with_failure()` to insure we aren't removing a correlation token when the operation is cleaning up due to a duplicate correlation token error. I'll trace the logic and see if we can't combine it and work through your scenario.

PR #402, opened 2025-06-18
Log ack_timeout_seconds

sbSteveK:
ack_timeout_seconds was skipped during the logging of MQTT5's client options storage. It has been added.

### DIFF
--- a/source/v5/mqtt5_options_storage.c
+++ b/source/v5/mqtt5_options_storage.c
@@ -3705,6 +3705,14 @@ void aws_mqtt5_client_options_storage_log(
         log_handle,
         level,
         AWS_LS_MQTT5_GENERAL,
+        "id=%p: aws_mqtt5_client_options_storage QoS 1 packet ack timeout interval set to %" PRIu32 " ms",
+        (void *)options_storage,
+        options_storage->ack_timeout_seconds);
+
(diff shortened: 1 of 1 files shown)


source/v5/mqtt5_options_storage.c
```diff
@@ -3701,6 +3701,14 @@ void aws_mqtt5_client_options_storage_log(
         (void *)options_storage,
         options_storage->connack_timeout_ms);
 
+    AWS_LOGUF(
+        log_handle,
+        level,
+        AWS_LS_MQTT5_GENERAL,
+        "id=%p: aws_mqtt5_client_options_storage QoS 1 packet ack timeout interval set to %" PRIu32 " ms",
```
sfod: Typo: `ms` instead of `s`
`to %" PRIu32 " ms"` ==> `to %" PRIu32 " s"`

sfod (approved):
Fix and ship

source/v5/mqtt5_options_storage.c
sbSteveK: good catch

xiazhvera: It's a little late, but does the timeout also applied to subscribe/unsubscribe operations, not only to QoS 1 operations?

PR #404, opened 2025-07-14
Auto-generate client ids in MQTT5

bretambrose:
IoT Core creates an invalid (cannot reconnect as) client id when the client uses MQTT5's auto-assigned client id feature.  This creates support problems for us since the client cannot reconnect when it should be able to.  This PR adds some logic to crudely guess at whether we're using IoT Core (is IoT Core validation turned on, which is the default for builder-created SDK clients) and manually creates a simple client ID in that case, avoiding the reconnect problem.


sbSteveK (approved):

source/v5/mqtt5_options_storage.c
```diff
@@ -3870,8 +3870,70 @@ struct aws_mqtt5_client_options_storage *aws_mqtt5_client_options_storage_new(
         options_storage->topic_aliasing_options = *options->topic_aliasing_options;
     }
 
+    struct aws_byte_buf auto_assign_id_buf;
+    AWS_ZERO_STRUCT(auto_assign_id_buf);
+
+    struct aws_mqtt5_packet_connect_view connect_options = *options->connect_options;
+    if (connect_options.client_id.len == 0) {
+        if (options->extended_validation_and_flow_control_options == AWS_MQTT5_EVAFCO_AWS_IOT_CORE_DEFAULTS) {
+            /*
```
xiazhvera, line 3879: We should update the comments for `AWS_MQTT5_EVAFCO_AWS_IOT_CORE_DEFAULTS` to mention that the client will generate the client ID. Other than that, the changes look good to me.

xiazhvera (approved):

PR #412, opened 2025-11-21
Add Platform Metrics Support For Mqtt3 Client

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Username field update: Now username field in MQTT CONNECT packet will treat string after the last "?" as URL query, and will re-process the string to form a valid query string. 
- Added a metrics struct
```
struct aws_mqtt_iot_sdk_metrics {
     struct aws_mqtt_metadata_entry *metadata_entries;
     size_t metadata_count;
    struct aws_byte_cursor library_name;
}
```

- Add API `aws_mqtt_client_connection_set_metrics` to set metrics for mqtt3. 
- Added Platform Metrics. Set platform metrics depends on the result returned by function  [struct aws_byte_cursor aws_get_platform_build_os_string(void)](https://github.com/awslabs/aws-c-common/blob/95515a8b1ff40d5bb14f965ca4cbbe99ad1843df/source/system_info.c#L99)

**SDK name field**
For the SDK name field, the library will use the format `IoTDeviceSDK/[Language]`. And the value would be unified across the CRT/SDKV2/SDKV1 for the same language.  e.x.: aws-crt-cpp/aws-iot-device-sdk-cpp/aws-iot-deivce-sdk-cpp-v2 will all be tracked as `IoTDeviceSDK/CPP`. The binding layer is responsible for setting the SDK name field. If not set, the name will default to be `IoTDeviceSDK/C`, indicating the client is using aws-c-mqtt library. 

**Username encoding**
IoT gateway team mentioned they are follow the query string format for username field. ( More specific, they used QueryStringDecoder API to decode the string. )  Beyond that, the username should follow the standard MQTT specs. 


source/client_impl_shared.c
```diff
@@ -227,3 +227,128 @@ bool aws_mqtt_byte_cursor_hash_equality(const void *a, const void *b) {
...
+    //         goto metrics_storage_error;
+    //     }
+
+    //     if (aws_array_list_push_back(&metrics_storage->metadata_entries, &entry)) {
+    //         goto metrics_storage_error;
+    //     }
+    // }
+
+    // storage_view->metadata_entries = metrics_storage->metadata_entries.data;
+    // storage_view->metadata_count = aws_array_list_length(&metrics_storage->metadata_entries);
```
sfod: If this commented really needed, it'd be better to hide it behind a feature flag, something like:

```c
// TODO ...
bool ENABLE_METADATA_METRICS = false;
if (ENABLE_METADATA_METRICS) {
   if (aws_array_list_init_dynamic(
   ...
}
```

This way, we'll be sure that this code is at least compilable.
Or, as alternative, extract it to a separate PR.

source/client.c
```diff
@@ -3341,6 +3373,46 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+static int s_aws_mqtt_client_connection_311_set_metrics(void *impl, const struct aws_mqtt_iot_sdk_metrics *metrics) {
+
+    struct aws_mqtt_client_connection_311_impl *connection = impl;
+
+    AWS_PRECONDITION(connection);
+    if (s_check_connection_state_for_configuration(connection)) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    if (metrics != NULL && aws_mqtt_validate_iot_sdk_metrics_utf8(metrics) == AWS_OP_ERR) {
```
sfod: We should add a sanity validation for a length. Some upper bound value that will be enough for everything. We can increase this limit in the future if needed.

sfod: Also, should we disallow some characters, e.g. `=`?

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <aws/common/encoding.h>
+#include <aws/common/string.h>
```
sfod: nit: unused includes

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+            goto error_output;
+        }
+    }
+
+    aws_byte_buf_clean_up(&metrics_string);
+    return AWS_OP_SUCCESS;
+
+error_output:
+    aws_byte_buf_clean_up(output_username);
+error:
```
sfod: nit: We shifted towards using one goto label and a cleanup function that can handle partially initialized things.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+#include <aws/common/system_info.h>
+#include <aws/mqtt/mqtt.h>
+#include <aws/mqtt/private/client_impl_shared.h>
+
+#include <stdio.h>
+
+// Maximum MQTT5 Content Type size https://docs.aws.amazon.com/general/latest/gr/iot-core.html#thing-limits
+const int AWS_IOT_MAX_CONTENT_SIZE = 256;
+
+int aws_mqtt_append_sdk_metrics_to_username(
```
sfod, line 98: Debatable: Since `username` may be altered by users, I think this function should be more resistant to various inputs.

For instance, if `username` has something like `MySDK=1.28.3`, then we won't add the SDK parameter because `has_sdk` will be true (and the same for `Platform`).

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+    aws_byte_buf_clean_up(&metrics_string);
+    return AWS_OP_ERR;
+}
+
+size_t aws_mqtt_append_sdk_metrics_to_username_size(
```
sfod: `aws_mqtt_append_sdk_metrics_to_username` and `aws_mqtt_append_sdk_metrics_to_username_size` share a lot of logic. Maybe try to reuse logic between them?

source/mqtt_iot_metrics.c
xiazhvera: The `SDK` and `Platform` attribute is reserved keywords for AWS IoT metrics collection, which means, we must fill `SDK` and `Platform` field so we could collect metrics properly. 

For other user customized attributes, like the case "MySDK=1.28.8", we will leave it there.

source/client_impl_shared.c
xiazhvera: Thinking about it, I'd better remove the metadata related comments.

include/aws/mqtt/private/client_impl_shared.h
```diff
@@ -132,6 +134,28 @@ struct aws_mqtt_client_connection {
     void *impl;
 };
 
+struct aws_mqtt_iot_sdk_metrics_storage {
```
bretambrose: Probably better off in a separate file.  this file is really intended for 311 (or the adapter)

include/aws/mqtt/private/client_impl_shared.h
```diff
@@ -144,4 +168,38 @@ AWS_MQTT_API bool aws_mqtt_byte_cursor_hash_equality(const void *a, const void *
 AWS_MQTT_API struct aws_event_loop *aws_mqtt_client_connection_get_event_loop(
     const struct aws_mqtt_client_connection *connection);
 
+/**
+ * Appends SDK metrics to the username field
```
bretambrose: Similar comment here, separate metrics file.

source/client.c
```diff
@@ -623,8 +623,32 @@ static void s_mqtt_client_init(
             &connect, topic_cur, connection->will.qos, connection->will.retain, payload_cur);
     }
 
-    if (connection->username) {
-        struct aws_byte_cursor username_cur = aws_byte_cursor_from_string(connection->username);
+    if (connection->username || connection->metrics) {
```
bretambrose: Not a fan of this conditional complexity.  We don't need to persist the username buffer in the connection object, we can just make a local buffer and free it after initing the packet.  The rest of the logic could just be internalized in a metrics API that takes (username, metrics, output_buffer) which you already have I think.

source/client.c
```diff
@@ -3341,6 +3373,46 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+
+    /* Clean up existing metrics if any */
+    if (connection->metrics) {
+        aws_mqtt_iot_sdk_metrics_storage_clean_up(connection->metrics);
+        connection->metrics = NULL;
+    }
+
+    if (metrics) {
+        /* Allocate metrics storage */
+        connection->metrics = aws_mem_calloc(connection->allocator, 1, sizeof(struct aws_mqtt_iot_sdk_metrics_storage));
```
bretambrose: If we only new/delete this object, let's use a new/delete api rather than an init/cleanup one that requires additional memory ops

source/client_impl_shared.c
```diff
@@ -227,3 +227,128 @@ bool aws_mqtt_byte_cursor_hash_equality(const void *a, const void *b) {
 struct aws_event_loop *aws_mqtt_client_connection_get_event_loop(const struct aws_mqtt_client_connection *connection) {
     return (*connection->vtable->get_event_loop)(connection->impl);
 }
+
+/*********************************************************************************************************************
+ * IoT SDK Metrics
+ ********************************************************************************************************************/
+
+size_t aws_mqtt_iot_sdk_metrics_compute_storage_size(const struct aws_mqtt_iot_sdk_metrics *metrics) {
```
bretambrose: Will this be used externally?

source/client_impl_shared.c
```diff
@@ -227,3 +227,128 @@ bool aws_mqtt_byte_cursor_hash_equality(const void *a, const void *b) {
...
+        metrics_storage->library_name = metrics_options->library_name;
+        if (aws_byte_buf_append_and_update(&metrics_storage->storage, &metrics_storage->library_name)) {
+            goto metrics_storage_error;
+        }
+        storage_view->library_name = metrics_storage->library_name;
+    }
+
+    return AWS_OP_SUCCESS;
+
+metrics_storage_error:
```
bretambrose: seems likely this is unreachable

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+#include <aws/common/encoding.h>
+#include <aws/common/string.h>
+#include <aws/common/system_info.h>
+#include <aws/mqtt/mqtt.h>
+#include <aws/mqtt/private/client_impl_shared.h>
+
+#include <stdio.h>
+
+// Maximum MQTT5 Content Type size https://docs.aws.amazon.com/general/latest/gr/iot-core.html#thing-limits
+const int AWS_IOT_MAX_CONTENT_SIZE = 256;
```
bretambrose: Why is this relevant?

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+
+int aws_mqtt_append_sdk_metrics_to_username(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *original_username,
+    const struct aws_mqtt_iot_sdk_metrics metrics,
```
bretambrose: Why is this pass-by-value?  You could eliminate that caller complexity by making this a pointer and on null returning the base username or ""

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    if (aws_mqtt_validate_iot_sdk_metrics_utf8(&metrics)) {
+        return AWS_OP_ERR;
+    }
+
+    /* Build metrics string */
+    struct aws_byte_buf metrics_string;
+    if (aws_byte_buf_init(&metrics_string, allocator, AWS_IOT_MAX_CONTENT_SIZE)) {
```
bretambrose: What makes us think that content type's limit is also applied to username?  If it actually is, we need to explicitly document that in comments.  Otherwise, this looks like a mistake.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /* Add SDK if not present */
+    if (!has_sdk) {
+        if (aws_byte_buf_append(&metrics_string, &sdk_str)) {
+            goto error;
+        }
+
+        struct aws_byte_cursor sdk_attr_value =
+            metrics.library_name.len > 0 ? metrics.library_name : aws_byte_cursor_from_c_str("IoTDeviceSDK/C");
```
bretambrose: Do we want to know the version # of aws-c-mqtt?

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+        if (aws_byte_buf_append(&metrics_string, &sdk_attr_value)) {
+            goto error;
+        }
+    }
+
+    /* Add Platform if not present */
+    if (!has_platform) {
+        struct aws_byte_cursor platform_cursor = aws_get_platform_build_os_string();
```
bretambrose: I don't like this function at all.  I'd much rather see an append platform string function that could be dynamic (ie include version and other dynamic information) rather than something that's just a handful of static strings.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,187 @@
...
+    if (aws_byte_buf_init(&metrics_string, allocator, AWS_IOT_MAX_CONTENT_SIZE)) {
+        return AWS_OP_ERR;
+    }
+
+    /* Check if the attributes already exists in username */
+    bool has_sdk = false;
+    bool has_platform = false;
+    bool has_query = false;
+
+    struct aws_byte_cursor sdk_str = aws_byte_cursor_from_c_str("SDK=");
```
bretambrose: As a personal preference I feel like we might be better served by some general purpose APIs for parsing and rebuilding query strings, similar to how URI implementations work, ie to change some internal element you parse the whole into the foundational pieces, modify them, and then regenerate the whole string from the parse.  That's in contrast to a lot of conditional logic here.  I think that would also be much nicer in the more general situation where there's metadata pairs and we're checking uniqueness, etc...

source/mqtt_iot_sdk_metrics.c
xiazhvera: Debatable: as we have a username storage, I think it is redundant to do a memcopy and return the base username here.

source/mqtt_iot_sdk_metrics.c
xiazhvera: Currently, I plan to put the version into metadata list, in a format of Metadata=(aws-c-mqtt=X.Y.Z, aws-crt-cpp=A.B.C), so that we could collect version numbers for all stacks.

include/aws/mqtt/private/client_impl_shared.h
sbSteveK: If this is the case, should we rename this to be more reflective of its use/purpose?

Looking further through this and its partner PR for mqtt5, it makes a lot of sense to put all this into its own file. Esp due to possibility of expansion of metrics related API as we expand its capabilities.

source/client_impl_shared.c
xiazhvera: No to the original question. Regardless, I combined `aws_mqtt_append_sdk_metrics_to_username` and `aws_mqtt_iot_sdk_metrics_compute_storage_size` into one function.

source/client.c
xiazhvera: I don't think "=" is disallowed, the user is allowed to add their own query which includes "=" sign. 
The length validation will be applied to the final username string.

source/mqtt_iot_sdk_metrics.c
xiazhvera: I updated the function so it build username from flexible query params, which user would have ability to pass in dynamic information. 
While I didn't see how we could avoid having the static string, as long as we would like to append the platform string by default.

source/client.c
xiazhvera: I removed the persist final username buffer and use a local one, while I didnt see much I can do with the conditional complexity here.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
```diff
@@ -0,0 +1,58 @@
...
+AWS_MQTT_API struct aws_mqtt_iot_sdk_metrics_storage *aws_mqtt_iot_sdk_metrics_storage_new(
+    struct aws_allocator *allocator,
+    const struct aws_mqtt_iot_sdk_metrics *metrics_options);
+
+AWS_MQTT_API size_t aws_mqtt_iot_sdk_metrics_compute_storage_size(const struct aws_mqtt_iot_sdk_metrics *metrics);
+
+AWS_MQTT_API void aws_mqtt_iot_sdk_metrics_storage_destroy(struct aws_mqtt_iot_sdk_metrics_storage *metrics_storage);
+
+/**
+ * Appends SDK metrics to the username field 1Code has comments. Press enter to view.
```
sfod: nit: something crept into the description

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    if (!metrics) {
+        if (out_full_username_size) {
+            *out_full_username_size = original_username->len;
+        }
+
+        if (output_username &&
+            aws_byte_buf_init(output_username, allocator, original_username->len) == AWS_OP_SUCCESS) {
+            aws_byte_buf_write(output_username, original_username->ptr, original_username->len);
```
sfod: nit: can be replaced with `aws_byte_buf_init_copy_from_cursor`. This should also help fix the case when on `aws_byte_buf_init` failure function returns `AWS_OP_SUCCESS`, but `output_username` is empty.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+        goto cleanup;
+    }
+
+    // build final output username
+    if (s_build_username_query(
+            original_username, base_username_length, &params_list, output_username, out_full_username_size)) {
+        goto cleanup;
+    }
+
+    aws_byte_buf_clean_up(&metrics_string);
```
sfod: nit: this is done in the `cleanup` section

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+        return AWS_OP_SUCCESS;
+    }
+
+    if (aws_mqtt_validate_iot_sdk_metrics_utf8(metrics)) {
+        return AWS_OP_ERR;
+    }
+
+    /* Build metrics string */
+    struct aws_byte_buf metrics_string;
+    if (aws_byte_buf_init(&metrics_string, allocator, AWS_IOT_MAX_USERNAME_SIZE)) {
```
sfod: Unconditionally allocating extra 128 KB seems excess.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *original_username,
+    const struct aws_mqtt_iot_sdk_metrics *metrics,
+    struct aws_byte_buf *output_username,
+    size_t *out_full_username_size) {
+
+    if (!allocator) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
```
sfod, line 110: Add assert for `original_username` is not NULL

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+    // The length of the base username part not including query parameters
+    size_t base_username_length = 0;
+    struct aws_byte_cursor question_mark_str = aws_byte_cursor_from_c_str("?");
+    struct aws_byte_cursor sdk_str = aws_byte_cursor_from_c_str("SDK");
+    struct aws_byte_cursor platform_str = aws_byte_cursor_from_c_str("Platform");
+
+    struct aws_array_list params_list;
+    aws_array_list_init_dynamic(&params_list, allocator, DEFAULT_QUERY_PARAM_COUNT, sizeof(struct aws_uri_param));
+
+    if (original_username && original_username->len > 0) {
```
sfod: `original_username` is used unconditionally above, so there is no point in checking it here. Assert at the beginning of the functions should cover this.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+#include <aws/common/byte_buf.h>
+#include <aws/common/system_info.h>
+#include <aws/common/uri.h>
+#include <aws/mqtt/mqtt.h>
+#include <aws/mqtt/private/mqtt_iot_sdk_metrics.h>
+
+#include <stdio.h>
+
+// MQTT payload size https://docs.aws.amazon.com/general/latest/gr/iot-core.html#thing-limits
+const size_t AWS_IOT_MAX_USERNAME_SIZE = 128 * 1024;
```
sfod: Something to discuss.
In packet.c, we effectively [limit](https://github.com/awslabs/aws-c-mqtt/blob/v0.13.3/source/packets.c#L26) the username to 65535 bytes. To avoid confusion in the future, we need some consistency on limits.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+    // if (aws_array_list_is_valid(&metrics_storage->metadata_entries)) {
+    //     aws_array_list_clean_up(&metrics_storage->metadata_entries);
+    // }
+
+    if (aws_byte_buf_is_valid(&metrics_storage->storage)) {
+        aws_byte_buf_clean_up(&metrics_storage->storage);
+    }
+    if (metrics_options != NULL) {
+        aws_mem_release(allocator, metrics_storage);
+    }
```
sfod: `if (metrics_options != NULL)` check is not needed here

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+        goto cleanup_storage;
+    }
+
+    metrics_storage->allocator = allocator;
+
+    struct aws_mqtt_iot_sdk_metrics *storage_view = &metrics_storage->storage_view;
+
+    // TODO Future Work: add metadata entries once we implemented the metadata feature
+    //
+    // if (aws_array_list_init_dynamic(
```
sfod: I still believe that we shouldn't commit commented code

source/client.c
```diff
@@ -3341,6 +3377,39 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+    struct aws_mqtt_client_connection_311_impl *connection = impl;
+
+    AWS_PRECONDITION(connection);
+    if (s_check_connection_state_for_configuration(connection)) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    if (metrics != NULL && aws_mqtt_validate_iot_sdk_metrics_utf8(metrics) == AWS_OP_ERR) {
+        AWS_LOGF_DEBUG(
+            AWS_LS_MQTT_CLIENT, "id=%p: Invalid utf8 or forbidden codepoints in metrics.", (void *)connection);
```
sfod: Maybe log as `ERROR` or `WARN`?

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,326 @@
...
+        goto cleanup;
+    }
+
+    if (output_username && aws_byte_buf_init(output_username, allocator, total_size)) {
+        goto cleanup;
+    }
+
+    // build final output username
+    if (s_build_username_query(
+            original_username, base_username_length, &params_list, output_username, out_full_username_size)) {
```
sfod: Restoring the username from params may slightly change it (e.g. `?a&b` becomes `?a=&b=`). I don't think it'll cause any issues, but we still should be aware of this.

source/client.c
```diff
@@ -3341,6 +3377,39 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+
+    /* Clean up existing metrics if any */
+    if (connection->metrics_storage) {
+        aws_mqtt_iot_sdk_metrics_storage_destroy(connection->metrics_storage);
+        connection->metrics_storage = NULL;
+    }
+
+    if (metrics) {
+        /* Initialize metrics storage */
+        connection->metrics_storage = aws_mqtt_iot_sdk_metrics_storage_new(connection->allocator, metrics);
```
sfod: `aws_mqtt_iot_sdk_metrics_storage_new` can fail. Should we return ERR in that case?

source/mqtt_iot_metrics.c
xiazhvera: Thinking about it, the function should handle the case where the original_username is a nullptr.

source/mqtt_iot_sdk_metrics.c
xiazhvera: Nice catch, updated s_build_username_query to handle the case. 

While I figured that this is a behavior change. As we didn't have any restriction on the username, while now we start to treat the username as query string. 
In a way, this should be fine, as the service side already treated the username as a query string, and use QueryStringDecoder to handle the username. It should build the same string we build here regardless. While, I will bring this to the team for further discussion.

source/mqtt_iot_sdk_metrics.c
xiazhvera: Update to use UINT16_MAX as size limit.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
```diff
@@ -0,0 +1,58 @@
...
+#ifndef AWS_MQTT_IOT_SDK_METRICS_H
+#define AWS_MQTT_IOT_SDK_METRICS_H
+
+/* IoT SDK Metrics */
+struct aws_mqtt_iot_sdk_metrics_storage {
+    struct aws_allocator *allocator;
+
+    struct aws_mqtt_iot_sdk_metrics storage_view;
+
+    struct aws_array_list metadata_entries;
```
sbSteveK: Trivial: Can we document/comment a little bit here on metadata_entries and library_name?

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
```diff
@@ -0,0 +1,58 @@
...
+    struct aws_byte_cursor library_name;
+
+    struct aws_byte_buf storage;
+};
+
+AWS_MQTT_API struct aws_mqtt_iot_sdk_metrics_storage *aws_mqtt_iot_sdk_metrics_storage_new(
+    struct aws_allocator *allocator,
+    const struct aws_mqtt_iot_sdk_metrics *metrics_options);
+
+AWS_MQTT_API size_t aws_mqtt_iot_sdk_metrics_compute_storage_size(const struct aws_mqtt_iot_sdk_metrics *metrics);
```
sbSteveK: If this is only used from within `mqtt_iot_sdk_metrics.c` we should probably not expose it as API. I think `aws_mqtt_append_sdk_metrics_to_username` is what's used outside of the metrics to determine storage size.

source/client.c
```diff
@@ -623,8 +626,25 @@ static void s_mqtt_client_init(
...
+        if (aws_mqtt_append_sdk_metrics_to_username(
+                connection->allocator,
+                &username_cur,
+                connection->metrics_storage ? &connection->metrics_storage->storage_view : NULL,
+                &username_with_metrics_buf,
+                NULL) == AWS_OP_SUCCESS) {
+            username_cur = aws_byte_cursor_from_buf(&username_with_metrics_buf);
+        } else {
+            AWS_LOGF_WARN(
+                AWS_LS_MQTT_CLIENT, "id=%p: Failed to apply metrics to username, using original", (void *)connection);
```
sbSteveK: It's possible that Username was NULL and the `aws_mqtt_append_sdk_metrics_to_username` failed. This Warning should probably only state we are using original if an original username existed.

Along that same line of thinking, if there was no username and metrics failed, we don't need to do anything below in this if block including processing of a password and the `aws_mqtt_packet_connect_add_credentials` call.

tests/CMakeLists.txt
```diff
@@ -133,6 +138,14 @@ add_test_case(mqtt5_shared_subscription_validation)
 # utf8 utility
 add_test_case(mqtt_utf8_encoded_string_test)
 
+# metrics utility
```
sbSteveK, line 144: If we are handling the possibility of a username being changed, we should test that the metrics on a follow-up connect is using the new username and generating the correct metrics.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#ifndef AWS_MQTT_IOT_SDK_METRICS_H
+#define AWS_MQTT_IOT_SDK_METRICS_H
+
+/* IoT SDK Metrics */
+struct aws_mqtt_iot_sdk_metrics_storage {
+    struct aws_allocator *allocator;
+
+    struct aws_mqtt_iot_sdk_metrics storage_view;
```
sbSteveK: We need a glossary or bible for how we use various terms. I think `view` indicates that the members within that need protection from being manipulated directly are const. The `aws_mqtt_iot_sdk_metrics` struct appears to only contain a byte cursor with a `library_name`. Let's just use the `aws_mqtt_iot_sdk_metrics_storage` directly and remove this `storage_view` ?

xiazhvera: I explored whether we could reduce complexity by removing the `view` concept from metrics storage. However, I think that keeping view in storage is actually simpler, as it allows all our functions to directly access view for operations. Without it, we would need separate APIs to handle view and storage independently but with duplicate functions. 

Regarding why we're not using const values in the metrics view: this is a design pain point in our current view design.
We use pointers to indicate optional values and direct values to indicate mandatory ones. For mandatory values, we cannot make them const in the view because we still need to allow modification of the underlying data when the storage gets updated.
You can see the same situation in our other view structs: https://github.com/awslabs/aws-c-mqtt/blob/main/include/aws/mqtt/v5/mqtt5_types.h#L334

For these reason, I will keep the `view` in metrics storage.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
xiazhvera: As this is the memory storage of the metrics, we'll document details at `aws_mqtt_iot_sdk_metrics_storage`.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
```diff
@@ -0,0 +1,56 @@
...
+    const struct aws_mqtt_iot_sdk_metrics *metrics_options);
+
+AWS_MQTT_API void aws_mqtt_iot_sdk_metrics_storage_destroy(struct aws_mqtt_iot_sdk_metrics_storage *metrics_storage);
+
+/**
+ * Appends SDK metrics to the username
+ *
+ * @param original_username The original username
+ * @param metrics The metrics configuration
+ * @param output_username Buffer to store the modified username. If the function succeed, caller is responsible to
```
bretambrose: Would add that this expects an uninitialized byte buf

include/aws/mqtt/client.h
```diff
@@ -693,6 +693,21 @@ int aws_mqtt_client_connection_get_stats(
...
+ * NOTE: DO NOT USE METADATA. Metadata will not be set.
+ *
+ * \param connection The connection object
+ * \param metrics The IoT SDK metrics configuration (pass NULL to disable metrics)
+ * \returns AWS_OP_SUCCESS if successful, AWS_OP_ERR otherwise
+ */
+AWS_MQTT_API
+int aws_mqtt_client_connection_set_metrics(
+    struct aws_mqtt_client_connection *connection,
+    const struct aws_mqtt_iot_sdk_metrics *metrics);
```
bretambrose: how is this getting forward declared?

include/aws/mqtt/mqtt.h
```diff
@@ -111,6 +111,44 @@ enum aws_mqtt_log_subject {
...
+struct aws_mqtt_iot_sdk_metrics {
+    /**
+     * TODO: This is a place holder for metadata entries.
+     * The metadata_entries is not added to username yet.
+     *
+     * DO NOT USE.
+     *
+     * Array of metadata entries
+     */
+    // struct aws_mqtt_metadata_entry *metadata_entries;
```
bretambrose: Can we delete or uncomment?

source/client.c
```diff
@@ -503,6 +504,8 @@ static void s_mqtt_client_init(
     AWS_FATAL_ASSERT((error_code != 0) == (channel == NULL));
 
     struct aws_mqtt_client_connection_311_impl *connection = user_data;
+    struct aws_byte_buf username_with_metrics_buf;
+    aws_byte_buf_init(&username_with_metrics_buf, connection->allocator, 0);
```
bretambrose: Why do this rather than zeroing?  I think this confuses the contract that the function is responsible for initializing the buffer.

source/client.c
```diff
@@ -623,25 +626,44 @@ static void s_mqtt_client_init(
...
+        } else {
+            AWS_LOGF_WARN(
+                AWS_LS_MQTT_CLIENT, "id=%p: Failed to apply metrics to username, using original", (void *)connection);
+        }
 
-        struct aws_byte_cursor password_cur = {
-            .ptr = NULL,
-            .len = 0,
-        };
+        if (aws_byte_cursor_is_valid(&username_cur)) {
```
bretambrose: Under what conditions would this not be valid?  Does it make sense to keep going?

source/client.c
```diff
@@ -623,25 +626,44 @@ static void s_mqtt_client_init(
...
+            AWS_LOGF_DEBUG(
+                AWS_LS_MQTT_CLIENT,
+                "id=%p: Adding username " PRInSTR " to connection",
+                (void *)connection,
+                AWS_BYTE_CURSOR_PRI(username_cur));
 
-        if (connection->password) {
-            password_cur = aws_byte_cursor_from_string(connection->password);
-        }
+            struct aws_byte_cursor password_cur = {
```
bretambrose: Could just zero struct

source/client.c
```diff
@@ -663,13 +685,21 @@ static void s_mqtt_client_init(
         goto handle_error;
     }
 
+    if (aws_byte_buf_is_valid(&username_with_metrics_buf)) {
```
bretambrose: Not sure this check is necessary.  If it is necessary then that may indicate some problems in the API contract.

source/client.c
```diff
@@ -3341,6 +3377,50 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
     return AWS_OP_SUCCESS;
 }
 
+static int s_aws_mqtt_client_connection_311_set_metrics(void *impl, const struct aws_mqtt_iot_sdk_metrics *metrics) {
+
+    struct aws_mqtt_client_connection_311_impl *connection = impl;
+
+    AWS_PRECONDITION(connection);
+    if (s_check_connection_state_for_configuration(connection)) {
```
bretambrose: I know you're just following the pattern but the pattern is broken.  This is completely unsafe; locks need to be held throughout the state update.

Something like:

```

storage = NULL;
if (metrics) {
   if (!validate_metrics()) {
      return aws_raise_error(AWS_ERROR_INVALID_ARG);
   }

   storage = aws_mqtt_iot_sdk_metrics_storage_new(connection->allocator, metrics);
   if (!storage) {
      return AWS_OP_ERR;
   }
}

int result = AWS_OP_ERR;
lock();

if (check_connection_state_fails()) {
   goto done;
}

connection->metrics_storage = storage;

storage = NULL; // keeps it from getting deleted
result = AWS_OP_SUCCESS;

done:

unlock();

if (storage) {
   destroy(storage);
}

return result;
```

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,293 @@
...
+    size_t base_username_length,
+    const struct aws_array_list *params_list,
+    struct aws_byte_buf *output_username,
+    size_t *out_full_username_size) {
+
+    AWS_ASSERT(base_username);
+    AWS_ASSERT(params_list);
+
+    if (output_username) {
+        aws_byte_buf_write(output_username, base_username->ptr, base_username_length);
```
bretambrose: This could fail?

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,293 @@
...
+    struct aws_byte_cursor query_param_amp = aws_byte_cursor_from_c_str("&");
+    struct aws_byte_cursor key_value_delim = aws_byte_cursor_from_c_str("=");
+
+    size_t params_count = aws_array_list_length(params_list);
+    for (size_t i = 0; i < params_count; ++i) {
+        struct aws_uri_param param;
+        AWS_ZERO_STRUCT(param);
+        aws_array_list_get_at(params_list, &param, i);
+
+        if (i == 0 && output_username) {
```
bretambrose: Compiler will do it for you but a logical simplification would be

```
if (output_username) {
    if (i == 0) {
       ...;
    } else {
       ...;
    }
}
```

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,293 @@
...
+        if (out_full_username_size) {
+            *out_full_username_size += 1;
+        }
+
+        if (output_username) {
+            if (param.key.len > 0 && (aws_byte_buf_append(output_username, &param.key))) {
+                return AWS_OP_ERR;
+            }
+
+            if (param.value.len > 0 && ((aws_byte_buf_append(output_username, &key_value_delim)) ||
```
bretambrose: This is logically correct but I don't like it as it's pretty deceptive

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    size_t storage_size = 0;
+
+    storage_size += metrics->library_name.len;
+
+    // TODO: add metadata entries when enabled
+    // for (size_t i = 0; i < metrics->metadata_count; ++i) {
+    //     storage_size += metrics->metadata_entries[i].key.len;
+    //     storage_size += metrics->metadata_entries[i].value.len;
```
bretambrose: conditionally add 1 based on value length

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,293 @@
...
+        aws_array_list_push_back(&params_list, &platform_params);
+    }
+
+    // Rebuild metrics string from params_list
+    // First parse to calculate total size
+    size_t total_size = 0;
+    s_build_username_query(&local_original_username, base_username_length, &params_list, NULL, &total_size);
+
+    if (total_size > AWS_IOT_MAX_USERNAME_SIZE) {
+        goto cleanup;
```
bretambrose, line 205: need to raise error and probably log here

include/aws/mqtt/private/mqtt_iot_metrics.h
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#ifndef AWS_MQTT_IOT_SDK_METRICS_H
+#define AWS_MQTT_IOT_SDK_METRICS_H
+
```
bretambrose, line 8: I don't understand how this is compiling without forward declarations

include/aws/mqtt/client.h
xiazhvera: Were you talking about `aws_mqtt_iot_sdk_metrics`? The struct is defined in `mqtt.h`, which is included in the file.

include/aws/mqtt/private/mqtt_iot_metrics.h
xiazhvera: Fixed by include "mqtt.h" here. 

I guess it passed the compiler as the `mqtt_iot_sdk_metrics.h` file always includes after `mqtt.h` in the code base.

source/client.c
xiazhvera: This is unlikely to happen, but the only scenario I can think of is when the username is not set, and the metrics string exceeds the size limit, resulting in an invalid username.

I believe it's still acceptable to continue the connection, as metrics are not mandatory for a successful connection.

sfod: > when the username is not set, and the metrics string exceeds the size limit

in that case, `username_cur` will be empty but still valid. I think we should verify the username is non-empty.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
sbSteveK: A bit confused. This is `aws_mqtt_iot_sdk_metrics_storage`. Does that mean we should be documenting details here?

`metadata_entries` is an `aws_array_list` and I can't infer what struct is going to be stored within this list. 

The issue might be that we don't currently use this in any way yet. Maybe it should also be commented out like all uses of it across this PR with the exception of in `aws_mqtt_iot_sdk_metrics_storage_destroy`.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
```diff
@@ -0,0 +1,59 @@
...
+AWS_MQTT_API
+int aws_mqtt_append_sdk_metrics_to_username(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *original_username,
+    const struct aws_mqtt_iot_sdk_metrics *metrics,
+    struct aws_byte_buf *output_username,
+    size_t *out_full_username_size);
+
+/**
+ * Validates that all string fields in aws_mqtt_iot_sdk_metrics
```
sbSteveK: Trivial: Remove "that" or add what "that" is referring to. e.g. "that all string fields in aws_mqtt_iot_sdk_metrics are valid"

include/aws/mqtt/client.h
```diff
@@ -693,6 +693,21 @@ int aws_mqtt_client_connection_get_stats(
     struct aws_mqtt_client_connection *connection,
     struct aws_mqtt_connection_operation_statistics *stats);
 
+/**
+ * Sets IoT SDK metrics configuration for the connection.
+ * These metrics will be appended to the username field during connection.
+ *
+ * NOTE: DO NOT USE METADATA. Metadata will not be set.
```
sbSteveK, line 700: This seems out of place. What is this referring to within the `aws_mqtt_client_connection_set_metrics` ?

include/aws/mqtt/mqtt.h
```diff
@@ -111,6 +111,24 @@ enum aws_mqtt_log_subject {
...
+ */
+struct aws_mqtt_metadata_entry {
+    struct aws_byte_cursor key;
+    struct aws_byte_cursor value;
+};
+
+/**
+ * IoT SDK metrics configuration structure
+ */
+struct aws_mqtt_iot_sdk_metrics {
```
sbSteveK: ooc, will this struct eventually be expanded to contain ALL metrics that are set from the SDK level? If that's the case are we planning on having a separate one for metrics from CRT? Depending on the future expansion or addition of structs we might want to reconsider naming this something more generic.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+
+    return AWS_OP_SUCCESS;
+}
+
+// TODO Future Work: we ignored the metadata field for now, will add them in future support
+int aws_mqtt_append_sdk_metrics_to_username(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *original_username,
+    const struct aws_mqtt_iot_sdk_metrics *metrics,
+    struct aws_byte_buf *output_username,
```
sbSteveK, line 102: Since this function expects the output_username `aws_byte_buf` to be uninitialized, should we add a check after the allocator NULL check that verifies this as well? Might be overkill but it explicitly sets the expectation here as well as in the definition for the function.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+    if (aws_array_list_is_valid(&metrics_storage->metadata_entries)) {
+        aws_array_list_clean_up(&metrics_storage->metadata_entries);
+    }
+    aws_byte_buf_clean_up(&metrics_storage->storage);
+
+    aws_mem_release(metrics_storage->allocator, metrics_storage);
+}
+
+int aws_mqtt_validate_iot_sdk_metrics(const struct aws_mqtt_iot_sdk_metrics *metrics) {
+    if (metrics == NULL) {
```
sbSteveK, line 297: Trivial behavior: Everywhere this function is called only calls this after checking whether metrics != NULL. I feel like metrics being NULL here should return an `AWS_OP_ERR` as the provided metrics are not valid.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+        return AWS_OP_SUCCESS;
+    }
+
+    if (aws_mqtt_validate_utf8_text(metrics->library_name)) {
+        return AWS_OP_ERR;
```
sbSteveK, line 302: We might want to log an error here. `aws_decode_utf8` returns AWS_OP_ERR but at no point logs the error or raises an appropriate error code.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+    // The length of the base username part not including query parameters
+    size_t base_username_length = 0;
+    struct aws_byte_cursor question_mark_str = aws_byte_cursor_from_c_str("?");
+    struct aws_byte_cursor sdk_str = aws_byte_cursor_from_c_str("SDK");
+    struct aws_byte_cursor platform_str = aws_byte_cursor_from_c_str("Platform");
+
+    struct aws_array_list params_list;
+    aws_array_list_init_dynamic(&params_list, allocator, DEFAULT_QUERY_PARAM_COUNT, sizeof(struct aws_uri_param));
+
+    if (local_original_username.len > 0) {
```
sbSteveK, line 141: Can we add a short description here regarding how metrics strings are constructed in relation to the `?` search? It'd help contextualize why we're looking for the last question mark below. This will aid in anyone coming to this in the future about why and how we're doing things. Especially if something is broken or we're digging for a bug.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+        struct aws_byte_cursor question_mark_find = local_original_username;
+
+        bool found_query = false;
+        // Find last question mark
+        while (AWS_OP_SUCCESS ==
```
sbSteveK, line 147: Might be worthwhile to also leave a comment about how we are finding the LAST `?` because that's what IoT does to collect this metric to preserve `?` that the user may have in their own username field.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+        // Find last question mark
+        while (AWS_OP_SUCCESS ==
+               aws_byte_cursor_find_exact(&question_mark_find, &question_mark_str, &question_mark_find)) {
+            // Advance cursor to skip the "?" character
+            aws_byte_cursor_advance(&question_mark_find, 1);
+            found_query = true;
+        }
+
+        if (found_query) {
+            // We dont want to keep the ? marker in the base username
```
sbSteveK: Trivial: We don't want to keep the metrics ? marker in the base username. User included ? might be kept?

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+               aws_byte_cursor_find_exact(&question_mark_find, &question_mark_str, &question_mark_find)) {
```
sbSteveK, line 148: Is it possible that we have a Username that was provided by the customer that has a `?` but we haven't added our own metrics to the base username? Would this result in us splitting the Customer username on their `?`?

tests/v3/connection_state_test.c
```diff
@@ -4010,3 +4010,206 @@ AWS_TEST_CASE_FIXTURE(
...
+    struct aws_mqtt_connection_options connection_options = {
+        .user_data = state_test_data,
+        .clean_session = false,
+        .client_id = aws_byte_cursor_from_c_str("client1234"),
+        .host_name = aws_byte_cursor_from_c_str(state_test_data->endpoint.address),
+        .socket_options = &state_test_data->socket_options,
+        .on_connection_complete = aws_test311_on_connection_complete_fn,
+    };
+
+    struct aws_byte_cursor username = aws_byte_cursor_from_c_str("testuser");
```
sbSteveK, line 4134: If it's valid for a username to have a `?` character as part of the username, let's add tests for the metrics that use a username like "test?user?".

source/client.c
```diff
@@ -623,25 +626,52 @@ static void s_mqtt_client_init(
...
+                    (void *)connection);
+            }
+        }
 
-        struct aws_byte_cursor password_cur = {
-            .ptr = NULL,
-            .len = 0,
-        };
+        if (aws_byte_cursor_is_valid(&username_cur)) {
+            AWS_LOGF_DEBUG(
```
sbSteveK: Do we want to log the entire metrics string? This may include metadata in the future.

Maybe we could log that we added username above where connection->username is checked and then another log that we added metrics to the username where we run `aws_mqtt_append_sdk_metrics_to_username`

include/aws/mqtt/private/client_impl.h
```diff
@@ -365,6 +365,11 @@ struct aws_mqtt_client_connection_311_impl {
      * Statistics tracking operational state
      */
     struct aws_mqtt_connection_operation_statistics_impl operation_statistics_impl;
+
+    /**
+     * IoT SDK metrics configuration
+     */
+    struct aws_mqtt_iot_sdk_metrics_storage *metrics_storage;
```
sbSteveK: Trivial: We are also logging from a direct from CRT level. Should we remove `sdk` from all metrics related functions/structs? e.g. `aws_mqtt_iot_metrics_storage`, `aws_mqtt_iot_metrics`, etc...

source/client.c
```diff
@@ -3341,6 +3380,67 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+static int s_aws_mqtt_client_connection_311_set_metrics(void *impl, const struct aws_mqtt_iot_sdk_metrics *metrics) {
+
+    struct aws_mqtt_client_connection_311_impl *connection = impl;
+    AWS_PRECONDITION(connection);
+
+    struct aws_mqtt_iot_sdk_metrics_storage *metrics_storage = NULL;
+    if (metrics) {
+        if (aws_mqtt_validate_iot_sdk_metrics(metrics) == AWS_OP_ERR) {
+            AWS_LOGF_ERROR(AWS_LS_MQTT_CLIENT, "id=%p: Invalid metrics.", (void *)connection);
+            return aws_raise_error(AWS_ERROR_INVALID_UTF8);
```
sbSteveK: This `AWS_ERROR_INVALID_UTF8` should be raised in the validation where it's failing. Here we should return `aws_last_error()`.

source/client.c
```diff
@@ -3341,6 +3380,67 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+        }
+
+        metrics_storage = aws_mqtt_iot_sdk_metrics_storage_new(connection->allocator, metrics);
```
sbSteveK: Trivial: Can we bring the declaration of `metrics_storage` down here? We could also combine the declaration and setting it to a single line if that makes things cleaner.

source/client.c
```diff
@@ -3341,6 +3380,67 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+            AWS_LS_MQTT_CLIENT,
+            "id=%p: Connection is currently pending connect/disconnect. Unable to make configuration changes until "
+            "pending operation completes.",
+            (void *)connection);
+        aws_raise_error(AWS_ERROR_INVALID_STATE);
+        goto done;
+    }
+
+    /* Clean up existing metrics if any */
+    if (connection->metrics_storage) {
```
sbSteveK: The first check in `aws_mqtt_iot_sdk_metrics_storage_destroy()` is a NULL check so we can skip this check and simply run the function on `connection->metrics_storage`

source/client.c
```diff
@@ -3341,6 +3380,67 @@ int aws_mqtt_client_connection_set_on_operation_statistics_handler(
...
+        aws_mqtt_iot_sdk_metrics_storage_destroy(connection->metrics_storage);
+    }
+    connection->metrics_storage = metrics_storage;
+    AWS_LOGF_TRACE(AWS_LS_MQTT_CLIENT, "id=%p: Setting IoT SDK metrics", (void *)connection);
```
sbSteveK, line 3623: Trivial: At this point we are not "Setting", it's already been set. Might reword the log or move this up to where we begin setting the metrics so that any errors logged happen after "Setting" has begun.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+#include <aws/mqtt/mqtt.h>
+#include <aws/mqtt/private/mqtt_iot_sdk_metrics.h>
+
+#include <stdio.h>
+
+// Use packet encoding limit for now: https://github.com/awslabs/aws-c-mqtt/blob/v0.13.3/source/packets.c#L26
+const size_t AWS_IOT_MAX_USERNAME_SIZE = UINT16_MAX;
+const size_t DEFAULT_QUERY_PARAM_COUNT = 10;
+
+// Build username query string from params_list, the caller is responsible to init and clean up output_username
```
sbSteveK: We should rewrite this to describe what's going on in this function.

It doesn't look like we are doing what's described on a NULL `output_username`. If we are doing that, we should add a NULL check against `output_username` early and set the `out_full_username_size` and exit the function instead of repeatedly checking it in the for loop.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+    size_t base_username_length,
+    const struct aws_array_list *params_list,
+    struct aws_byte_buf *output_username,
+    size_t *out_full_username_size) {
+
+    AWS_ASSERT(base_username);
+    AWS_ASSERT(params_list);
+
+    if (output_username) {
+        if (!aws_byte_buf_write(output_username, base_username->ptr, base_username_length)) {
```
sbSteveK, line 40: I don't understand why we're doing this check here. The `base_username_length` doesn't include the added length required for the params being added below. I think we should just fail in the appends below instead of adding this check here.

source/mqtt_iot_sdk_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+// If output_username is NULL, the function will just calculate the full username size and return it in
+// out_full_username_size
+int s_build_username_query(
+    const struct aws_byte_cursor *base_username,
+    size_t base_username_length,
+    const struct aws_array_list *params_list,
+    struct aws_byte_buf *output_username,
+    size_t *out_full_username_size) {
```
sbSteveK: I feel like the way this is used, we should have a separate function that gets the full username size instead of doing both the building of a username and its size in a shared function. Especially since this is called twice down below with the first call being just to get the size and only the second time generating the output_username.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+            .key = platform_str,
+            .value = aws_get_platform_build_os_string(),
+        };
+        aws_array_list_push_back(&params_list, &platform_params);
+    }
+
+    // Rebuild metrics string from params_list
+    // First parse to calculate total size
+    size_t total_size = 0;
+    s_build_username_query(&local_original_username, base_username_length, &params_list, NULL, &total_size);
```
sbSteveK, line 199: As mentioned above, this could call a specific metrics total size function with the `base_username_length` and `params_list` provided.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,299 @@
...
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        goto cleanup;
+    }
+
+    if (output_username && aws_byte_buf_init(output_username, allocator, total_size)) {
+        goto cleanup;
+    }
+
+    // build final output username
+    if (s_build_username_query(
```
sbSteveK, line 213: And this would just stay the same minus the size stuff that was already gotten up above. Feels off to call `s_build_username_query()` for the purpose of getting a total size and only a total size.

source/mqtt_iot_metrics.c
xiazhvera: The function `aws_mqtt_validate_utf8_text` will raise error `AWS_ERROR_INVALID_UTF8`. It is upon caller to decide if the client should log the error.

source/client.c
xiazhvera: This log is here to capture the final data being added to the connect packet for debugging purposes. I think it is valuable to have the complete string logged.

source/client.c
xiazhvera: Unfortunately, the `metrics_storage` will be used later for assignment, so it could not be declared in this scope.

source/mqtt_iot_sdk_metrics.c
xiazhvera: If we separate these into two functions, both would need to implement the same logic for determining how the final username is constructed (base username + delimiter + metrics). Any future changes to the format would require updating both functions, and they could easily drift out of sync.

source/mqtt_iot_metrics.c
xiazhvera: The same as above.

source/mqtt_iot_sdk_metrics.c
xiazhvera: We still need calculate the final output size instead of exist the funciton if `output_username` is NULL.

include/aws/mqtt/private/mqtt_iot_sdk_metrics.h
xiazhvera: Sorry I mean to say `aws_mqtt_iot_sdk_metrics`. As the `aws_mqtt_iot_sdk_metrics_storage_destroy ` is the memory storage for  `aws_mqtt_iot_sdk_metrics`, the details will be documented at `aws_mqtt_iot_sdk_metrics`. 

As `metadata` is not implemented, it is now removed the from this PR. However, once it's there, it will be an array list of `aws_mqtt_metadata_entry`, which is a struct of `aws_byte_cursor` key/value pair

source/mqtt_iot_metrics.c
xiazhvera: Technically yes, but it's highly unlikely in practice as the current IoT SDK metrics implementation already uses `?` as a query delimiter. Any existing user with `?` in their username would already be broken.

sbSteveK (approved):

PR #413, opened 2025-11-21
Add Metrics Support For Mqtt5 Client 

xiazhvera:
*Issue #, if available:*

*Description of changes:*


include/aws/mqtt/v5/mqtt5_types.h
```diff
@@ -394,6 +396,9 @@ struct aws_mqtt5_packet_connect_view {
     /* Do not bind these.  We don't support AUTH packets yet.  For decode/encade testing purposes only. */
     const struct aws_byte_cursor *authentication_method;
     const struct aws_byte_cursor *authentication_data;
+
+    /* IoT SDK metrics configuration */
+    const struct aws_mqtt_iot_sdk_metrics *metrics;
```
bretambrose: Let's not put this in the connect view.  Why not the client config?

bretambrose (changes requested):

include/aws/mqtt/private/v5/mqtt5_options_storage.h
```diff
@@ -139,6 +138,7 @@ struct aws_mqtt5_operation_pingreq {
 
 struct aws_mqtt5_client_options_storage {
     struct aws_allocator *allocator;
+    struct aws_mqtt5_client_options options;
```
sfod: This doesn't look right. Other fields in `aws_mqtt5_client_options_storage` mirror fields from  `aws_mqtt5_client_options`, so it should be similar with metrics.

source/v5/mqtt5_options_storage.c
```diff
@@ -3929,19 +3994,25 @@ struct aws_mqtt5_client_options_storage *aws_mqtt5_client_options_storage_new(
     }
```
sfod, line 3975: Can we modify `username` here and store the modified value in the connect storage? So, there will be this local change instead of adding mqtt5 options in multiple places where it adds unnecessary dependencies.

xiazhvera: Debatable, but here is my concerns: for one side, I would like to keep the original username in case we could extend it in the future. (For example, if we allow user to modify the username/metrics for reconnection.) 
Meanwhile our `aws_mqtt5_packet_connect_view` is a data structure represents the MQTT CONNECT packet, therefore, things like metrics should not live in connect packet. Then eventually I add it to the client option.

source/v5/mqtt5_options_storage.c
```diff
@@ -666,12 +676,13 @@ static size_t s_aws_mqtt5_packet_connect_compute_storage_size(const struct aws_m
 int aws_mqtt5_packet_connect_storage_init(
     struct aws_mqtt5_packet_connect_storage *storage,
     struct aws_allocator *allocator,
-    const struct aws_mqtt5_packet_connect_view *view) {
+    const struct aws_mqtt5_packet_connect_view *view,
+    const struct aws_mqtt5_client_options *client_config) {
```
sbSteveK: Trivial: Can we try to use the same naming as we're passing this around? It's `client_config` here but becomes `options` after passing it to `s_aws_mqtt5_packet_connect_compute_storage_size`

sbSteveK: Looking further through, maybe we should change them from `client_config` to `client_options` for clarity throughout.

include/aws/mqtt/private/v5/mqtt5_options_storage.h
```diff
@@ -217,7 +219,8 @@ AWS_MQTT_API uint32_t aws_mqtt5_operation_get_ack_timeout_override(const struct
 
 AWS_MQTT_API struct aws_mqtt5_operation_connect *aws_mqtt5_operation_connect_new(
```
sbSteveK, line 219: I feel that this should take the `aws_mqtt5_client_options_storage` and use its `metrics_storage` rather than creating an `aws_mqtt5_client_options` within the storage for sharing the metrics storage. We directly access storage values in numerous other places and should probably do so here unless there's a good reason not to.

My understanding is that we use view's when we are sharing things with the customer and don't want them to modify things in the storage that are being exposed. That doesn't feel like it's a concern here as this is all internal.

include/aws/mqtt/v5/mqtt5_packet_storage.h
```diff
@@ -209,7 +211,8 @@ AWS_MQTT_API size_t aws_mqtt5_user_property_set_size(const struct aws_mqtt5_user
 AWS_MQTT_API int aws_mqtt5_packet_connect_storage_init(
     struct aws_mqtt5_packet_connect_storage *connect_storage,
     struct aws_allocator *allocator,
-    const struct aws_mqtt5_packet_connect_view *connect_options);
+    const struct aws_mqtt5_packet_connect_view *connect_options,
```
bretambrose: Why both?  The connect is a member of the client options.

source/v5/mqtt5_options_storage.c
```diff
@@ -666,12 +680,13 @@ static size_t s_aws_mqtt5_packet_connect_compute_storage_size(const struct aws_m
 int aws_mqtt5_packet_connect_storage_init(
     struct aws_mqtt5_packet_connect_storage *storage,
     struct aws_allocator *allocator,
-    const struct aws_mqtt5_packet_connect_view *view) {
+    const struct aws_mqtt5_packet_connect_view *view,
+    const struct aws_mqtt5_client_options_storage *client_storage) {
```
bretambrose: client_options_storage or similar

source/v5/mqtt5_options_storage.c
```diff
@@ -686,9 +701,28 @@ int aws_mqtt5_packet_connect_storage_init(
...
+                    client_storage->metrics_storage ? &client_storage->metrics_storage->storage_view : NULL,
+                    &metrics_username_buf,
+                    NULL)) {
+                return AWS_OP_ERR;
+            }
+            storage->username = aws_byte_cursor_from_buf(&metrics_username_buf);
+        }
+
         if (aws_byte_buf_append_and_update(&storage->storage, &storage->username)) {
+            aws_byte_buf_clean_up(&metrics_username_buf);
```
bretambrose: Could avoid the need for two clean up calls with a 

```
int result = aws_byte_buf_append_and_update(&storage->storage, &storage->username);
aws_byte_buf_clean_up(&metrics_username_buf);
if (result) {
   return AWS_OP_ERR;
}
```

source/v5/mqtt5_options_storage.c
```diff
@@ -3743,6 +3786,10 @@ void aws_mqtt5_client_options_storage_destroy(struct aws_mqtt5_client_options_st
         aws_mem_release(options_storage->connect->allocator, options_storage->connect);
     }
 
+    if (options_storage->metrics_storage != NULL) {
```
bretambrose: I prefer the null checks in the destructor itself.  Makes usage simpler.

tests/v5/mqtt5_operation_and_storage_tests.c
```diff
@@ -689,14 +692,12 @@ static int s_mqtt5_connect_storage_new_set_all_fn(struct aws_allocator *allocato
         .user_property_count = AWS_ARRAY_SIZE(s_user_properties),
         .user_properties = s_user_properties,
         .authentication_method = &s_authentication_method_cursor,
-        .authentication_data = &s_authentication_data_cursor,
-    };
+        .authentication_data = &s_authentication_data_cursor};
```
bretambrose: Readd the comma for formatting

bretambrose (approved):
Fix n ship

PR #414, opened 2025-11-21
Add Metrics Support For Mqtt5-to-3 Adapter

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Added metrics support for Mqtt5-to-3 Adapter. Allow setting metrics on the adapter. 


tests/v5/mqtt5_to_mqtt3_adapter_tests.c
```diff
@@ -4406,3 +4406,118 @@ static int s_mqtt5to3_adapter_operation_callbacks_after_shutdown_fn(struct aws_a
...
+
+    ASSERT_SUCCESS(aws_mqtt_client_connection_set_metrics(connection, &metrics));
+
+    aws_mqtt5_to_mqtt3_adapter_test_fixture_clean_up(&fixture);
+    aws_mqtt_library_clean_up();
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(mqtt5to3_adapter_set_metrics_valid, s_mqtt5to3_adapter_connection_set_metrics_valid_fn)
```
sbSteveK, line 4443: It looks like these tests may not have been added to the CMakeLists.txt

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2779,6 +2779,99 @@ static int s_aws_mqtt_client_connection_5_get_stats(
...
+static void s_set_metrics_task_fn(struct aws_task *task, void *arg, enum aws_task_status status) {
+    (void)task;
+
+    struct aws_mqtt_set_metrics_task *set_task = arg;
+    struct aws_mqtt_client_connection_5_impl *adapter = set_task->adapter;
+    if (status != AWS_TASK_STATUS_RUN_READY) {
+        goto done;
+    }
+
+    if (set_task->metrics_storage && aws_mqtt_validate_iot_sdk_metrics(&set_task->metrics_storage->storage_view)) {
```
bretambrose: Why wouldn't we check this before creating the task?

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2779,6 +2779,99 @@ static int s_aws_mqtt_client_connection_5_get_stats(
...
+        goto done;
+    }
+
+    /* we're in the mqtt5 client's event loop; it's safe to access internal state */
+    struct aws_mqtt5_client_options_storage *client_options = adapter->client->config;
+    if (client_options->metrics_storage) {
+        aws_mqtt_iot_sdk_metrics_storage_destroy(client_options->metrics_storage);
+    }
+
+    client_options->metrics_storage = aws_mqtt_iot_sdk_metrics_storage_new(
```
bretambrose: Is this failable?  Maybe validate and create storage before hand, letting the task track the storage and then this function just destroys existing and assigns.

Actually, the task already has the storage, why not just "move" it by assigning the task value and zeroing it out so that it doesn't get destroyed?

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2779,6 +2779,99 @@ static int s_aws_mqtt_client_connection_5_get_stats(
...
+done:
+    aws_ref_count_release(&adapter->internal_refs);
+    s_aws_mqtt_set_metrics_task_destroy(set_task);
+}
+
+static struct aws_mqtt_set_metrics_task *s_aws_mqtt_set_metrics_task_new(
+    struct aws_allocator *allocator,
+    struct aws_mqtt_client_connection_5_impl *adapter,
+    const struct aws_mqtt_iot_sdk_metrics *metrics) {
+
```
bretambrose, line 2827: Validate here, also fail if new storage fails too

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2779,6 +2779,99 @@ static int s_aws_mqtt_client_connection_5_get_stats(
...
+    return set_task;
+}
+
+static int s_aws_mqtt_client_connection_5_set_metrics(void *impl, const struct aws_mqtt_iot_sdk_metrics *metrics) {
+    struct aws_mqtt_client_connection_5_impl *adapter = impl;
+
+    if (aws_mqtt_validate_iot_sdk_metrics(metrics)) {
+        AWS_LOGF_DEBUG(
+            AWS_LS_MQTT5_TO_MQTT3_ADAPTER, "id=%p: Invalid utf8 or forbidden codepoints in metrics.", (void *)adapter);
+        return aws_raise_error(AWS_ERROR_INVALID_UTF8);
```
bretambrose: This should be set in the validation function not here.  Eventually validation might include non-utf8 errors

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2779,6 +2779,99 @@ static int s_aws_mqtt_client_connection_5_get_stats(
...
+        set_task->metrics_storage = aws_mqtt_iot_sdk_metrics_storage_new(allocator, metrics);
+    }
+
+    return set_task;
+}
+
+static int s_aws_mqtt_client_connection_5_set_metrics(void *impl, const struct aws_mqtt_iot_sdk_metrics *metrics) {
+    struct aws_mqtt_client_connection_5_impl *adapter = impl;
+
+    if (aws_mqtt_validate_iot_sdk_metrics(metrics)) {
```
bretambrose: Ok, we are validating synchronously.  That means we don't need to validate in the task function right?

source/v5/mqtt5_to_mqtt3_adapter.c
```diff
@@ -2779,6 +2779,97 @@ static int s_aws_mqtt_client_connection_5_get_stats(
...
+    s_aws_mqtt_set_metrics_task_destroy(set_task);
+}
+
+static struct aws_mqtt_set_metrics_task *s_aws_mqtt_set_metrics_task_new(
+    struct aws_allocator *allocator,
+    struct aws_mqtt_client_connection_5_impl *adapter,
+    const struct aws_mqtt_iot_sdk_metrics *metrics) {
+
+    struct aws_mqtt_set_metrics_task *set_task = aws_mem_calloc(allocator, 1, sizeof(struct aws_mqtt_set_metrics_task));
+    if (set_task == NULL) {
```
sbSteveK: Trivial: Don't do NULL checks on `aws_mem_calloc`. We moved away from returning NULL to a fatal error when this goes bad.

sbSteveK (approved):

PR #417, opened 2026-02-03
Manual Puback for MQTT5

sbSteveK:
Implements manual PUBACK feature for the MQTT5 client.

Provides public API calls:
`aws_mqtt5_client_acquire_puback()`
`aws_mqtt5_client_invoke_puback()`

Tests for new API


include/aws/mqtt/manual-puback/manual_puback.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/mqtt/v5/mqtt5_client.h>
+
+/**
+ * Result type for manual PUBACK operations.  Lets the caller signal if the manual ack
+ * operation succeeded or if it failed (and why).
```
xiazhvera: nit: "Let the caller signal ... "

include/aws/mqtt/private/v5/mqtt5_client_impl.h
```diff
@@ -252,6 +252,15 @@ struct aws_mqtt5_client_operational_state {
      */
     aws_mqtt5_packet_id_t next_mqtt_packet_id;
 
+    /*
+     * One more than the most recently used control packet id.
+     */
+    uint64_t next_mqtt5_puback_control_id;
+    struct aws_hash_table manual_puback_packet_id_table;
```
xiazhvera, line 261: Probably add more explanation about the tables, includes what they do and what's the key-value types.

source/manual-puback/manual_puback.c
```diff
@@ -0,0 +1,252 @@
...
+    if (aws_hash_table_put(
+            &client->operational_state.manual_puback_packet_id_table, &manual_puback->packet_id, manual_puback, NULL)) {
+        int error_code = aws_last_error();
+        AWS_LOGF_ERROR(
+            AWS_LS_MQTT5_CLIENT,
+            "id=%p: Failed to insert manual PUBACK entry into packet ID table: %d(%s)",
+            (void *)client,
+            error_code,
+            aws_error_debug_str(error_code));
+        aws_mem_release(manual_puback->allocator, manual_puback);
```
xiazhvera: Debatable, but I prefer a single failure point for cleanup, which would be easier to manage and avoid memory leaks on the manual_puback.  For example, like what we did here : https://github.com/awslabs/aws-c-mqtt/blob/bc8ec4fdb678dbbb14a3062f110976a3ab1ceb8c/tests/v3/mqtt_mock_server_handler.c#L300

source/manual-puback/manual_puback.c
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    // The current_control_packet_id is incremented each time a new manual puback is scheduled.
+    uint64_t current_control_packet_id = client->operational_state.next_mqtt5_puback_control_id;
+    struct aws_mqtt5_manual_puback_entry *manual_puback =
+        s_aws_mqtt_manual_puback_entry_new(client->allocator, publish_view->packet_id, current_control_packet_id);
+
+    /* Allows lookup of manual puback entries by packet id */
+    if (aws_hash_table_put(
+            &client->operational_state.manual_puback_packet_id_table, &manual_puback->packet_id, manual_puback, NULL)) {
```
xiazhvera: Reading the code below, it seems that the packet_id_table should have `packet_id` as key and `control_id` as value instead of the `manual_puack` itself. 
It would be nice to document the key-value types/usage for the table in the header file.

include/aws/mqtt/manual-puback/manual_puback.h
```diff
@@ -0,0 +1,70 @@
+#ifndef AWS_MQTT_MANUAL_PUBACK_H
```
bretambrose: Wouldn't this make sense to be in v5 and not "manual-puback"

include/aws/mqtt/manual-puback/manual_puback.h
```diff
@@ -0,0 +1,70 @@
...
+#define AWS_MQTT_MANUAL_PUBACK_H
+
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/mqtt/v5/mqtt5_client.h>
```
bretambrose: remove in favor of forward declarations

include/aws/mqtt/manual-puback/manual_puback.h
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Takes manual control of PUBACK for the given PUBLISH packet.
+ *
+ * @param client mqtt5 client that received the PUBLISH packet to take manual PUBACK control from.
+ * @param publish_view the view of the PUBLISH packet that PUBACK control is taken from.
+ * @return puback_control_id of the PUBLISH packet. This can be used to schedule a PUBACK for the PUBLISH packet
+ * when used with the aws_mqtt5_manual_puback function call.
+ */
+AWS_MQTT_API uint64_t aws_mqtt5_take_manual_puback_control(
```
bretambrose: Feels like these declarations should just be in the public mqtt5 client file.  Impl can stay separate.

bretambrose: This is a client operation, so naming wise "aws_mqtt5_client_..."

Also I'd prefer "acquire" as the verb not "take".... "control"

So

`aws_mqtt5_client_acquire_puback`

`aws_mqtt5_client_invoke_puback` or `aws_mqtt5_client_release_puback`

include/aws/mqtt/private/manual-puback/manual_puback_impl.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * After a PUBLISH packet has been provided to the user, this check will determine if the PUBACK has had its PUBACK
+ * control taken. If the PUBACK control has been taken, the user is responsible for sending a PUBACK for the PUBLISH
+ * packet.
+ *
+ * @param client mqtt5 client that received the PUBLISH packet to take manual PUBACK control from.
+ * @param publish_view the view of the PUBLISH packet that's PUBACK is being handled.
+ */
+void aws_mqtt5_handle_puback(struct aws_mqtt5_client *client, const struct aws_mqtt5_packet_publish_view *publish_view);
```
bretambrose: TBH I don't understand this function's doc comment at all.

include/aws/mqtt/private/manual-puback/manual_puback_impl.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+/**
+ * This is used to track which PUBLISH packets a user has taken manual PUBACK control from.
+ */
+struct aws_mqtt5_manual_puback_entry {
```
bretambrose: Wondering why this needs to be in a header file

include/aws/mqtt/private/v5/mqtt5_client_impl.h
```diff
@@ -252,6 +252,15 @@ struct aws_mqtt5_client_operational_state {
      */
     aws_mqtt5_packet_id_t next_mqtt_packet_id;
 
+    /*
+     * One more than the most recently used control packet id.
+     */
```
bretambrose, line 257: My suggestion to wall off the implementation as a separate system was intended to encapsulate/insulate the client from this kind of change, ie the puback system is an opaque structure whose logic is completely divorced from the client.  The only crossover would be whatever function the impl needs in order to queue a puback.

Essentially, publish received would now be the responsibility of this new system.  It would have a clean, private API for use by the client and the client itself would have very few changes in it (just creating/destroying and invoking the system).

bretambrose (commented):
Looked at the headers

source/v5/mqtt5_client.c
```diff
@@ -3392,3 +3460,16 @@ bool aws_mqtt5_client_reset_connection(struct aws_mqtt5_client *client) {
 
     return false;
 }
+
+/* Wrapper functions to expose s_aws_mqtt5_client_queue_puback for manual PUBACK functionality */
```
xiazhvera: I didn't get why we wrap the function instead of directly make change to the original static function ...?

source/v5/mqtt5_client.c
```diff
@@ -590,6 +592,40 @@ static void s_enqueue_operation_front(struct aws_mqtt5_client *client, struct aw
...
+        AWS_LOGF_INFO(
+            AWS_LS_MQTT5_CLIENT,
+            "id=%p: Clearing %zu manual PUBACK control ids.",
+            (void *)client_operational_state->client,
+            count);
+        aws_hash_table_clear(&client_operational_state->manual_puback_packet_id_table);
+        aws_hash_table_foreach(
+            &client_operational_state->manual_puback_control_id_table,
+            s_manual_puback_transfer,
+            &client_operational_state->manual_puback_cancelled_control_id_table);
```
xiazhvera, line 635: While unlikely, we should still handle the case where `s_manual_puback_transfer` fails during `aws_hash_table_foreach` (returning AWS_COMMON_HASH_TABLE_ITER_ERROR). Unhandled error could eventually lead to a leak.

source/v5/mqtt5_client.c
```diff
@@ -590,6 +592,40 @@ static void s_enqueue_operation_front(struct aws_mqtt5_client *client, struct aw
     s_reevaluate_service_task(client);
 }
 
+/* This is used to move the aws_mqtt5_manual_puback_entry stored in the active table to cancelled table. We don't simply
+ * clear out the active table because we need to keep the entries in memory until we are fully done with them. */
+static int s_manual_puback_transfer(void *context, struct aws_hash_element *element) {
```
xiazhvera, line 595: In general, I'm not a fan of transferring memory ownership between tables. I'd propose making `manual_puback` ref-counted instead. On each `aws_hash_table_put`, we acquire a reference, and on removal, we release it. This way, each table independently manages its own reference lifetime.

include/aws/mqtt/manual-puback/manual_puback.h
sbSteveK: Moving the include to manual_puback.c and forward declaring here

include/aws/mqtt/manual-puback/manual_puback.h
sbSteveK: This is a remnant of initial implementation where we were considering MQTT3 support. I considered moving it into the v5 folder but felt like I'd rather keep it within mqtt/manual-puback in case we add it later. I'm fine with moving things into v5. Should be a minimal change of folder structure if we want it in v5 instead.

include/aws/mqtt/private/manual-puback/manual_puback_impl.h
sbSteveK: I was using this in both manual_puback.c and mqtt5_client.c but didn't update after I moved everything from mqtt5_client.c to manual_puback.c during removal of ordered pubacks. I've removed it from here and put it into manual_puback.c

include/aws/mqtt/manual-puback/manual_puback.h
sbSteveK: Changed it to simply "Result for manual PUBACK operations."

include/aws/mqtt/manual-puback/manual_puback.h
sbSteveK: Changed to `aws_mqtt5_client_acquire_puback` and `aws_mqtt5_client_invoke_puback`

sbSteveK: I don't think it needs to be all or nothing but the lines feel blurred between moving these functions into the public mqtt5 client and moving the implementation details out on their own. i.e. current_control_id. I'm beginning to feel like the entire manual puback should simply be a part of the mqtt5 client as it's being handled as a client operation ala Publish, Subscribe, etc.

source/v5/mqtt5_client.c
sbSteveK: I originally wanted to keep things as separate as possible between the client and the manual puback functions. I'm removing this separation and will clear these wrappers out.

include/aws/mqtt/private/manual-puback/manual_puback_impl.h
sbSteveK: This entire function has been removed with the migration into mqtt5_client

include/aws/mqtt/private/v5/mqtt5_client_impl.h
sbSteveK: done

source/manual-puback/manual_puback.c
sbSteveK: We need to use the allocated manual_puback as its value because we need to insure that the pointers used as the keys for both tables stay valid until we remove/destroy them.

source/manual-puback/manual_puback.c
sbSteveK: works for me

source/v5/mqtt5_client.c
sbSteveK: I think if this fails we've got bigger problems outside of a memory leak. Looked around and don't see anything that handles an error within the aws_hash_table_foreach() called function.

include/aws/mqtt/private/v5/mqtt5_options_storage.h
```diff
@@ -272,12 +274,18 @@ AWS_MQTT_API void aws_mqtt5_packet_publish_view_log(
...
     struct aws_allocator *allocator,
-    const struct aws_mqtt5_packet_puback_view *puback_options);
+    const struct aws_mqtt5_packet_puback_view *puback_options,
+    const struct aws_mqtt5_manual_puback_completion_options *completion_options);
 
 AWS_MQTT_API void aws_mqtt5_packet_puback_view_log(
     const struct aws_mqtt5_packet_puback_view *puback_view,
     enum aws_log_level level);
 
+AWS_MQTT_API struct aws_mqtt5_manual_puback_entry *s_aws_mqtt_manual_puback_entry_new(
```
sfod: This function is used only in `mqtt5_client.c`. Moving it there (and making it static) would be cleaner.

include/aws/mqtt/v5/mqtt5_client.h
```diff
@@ -738,6 +755,32 @@ int aws_mqtt5_client_publish(
...
+    struct aws_mqtt5_client *client,
+    const struct aws_mqtt5_packet_publish_view *publish_view);
+
+/**
+ * Send PUBACK for provided control id. Callback in completion_options will be invoked with an
+ * aws_mqtt5_manual_puback_result once a PUBACK operation has been completed.
+ *
+ * @param client mqtt5 client to queue a puback for
+ * @param puback_control_id Control ID of aws_mqtt5_manual_puback_entry to send to broker/server
+ * @return success/failure of the manual PUBACK operation.
```
sfod: nit: success/failure of starting the PUBACK operation

include/aws/mqtt/v5/mqtt5_client.h
```diff
@@ -294,6 +294,15 @@ typedef void(aws_mqtt5_publish_completion_fn)(
...
     void *complete_ctx);
 
+/**
+ * Signature of callback invoked when a manual PUBACK operation completes.
+ *
+ * @param puback_result result of the PUBACK operation
+ * @param complete_ctx user data passed in with the completion options
+ */
+typedef void(
+    aws_mqtt5_manual_puback_completion_fn)(enum aws_mqtt5_manual_puback_result puback_result, void *complete_ctx);
```
sfod: nit: Maybe rename `complete_ctx` into `completion_user_data`? For naming consistency

source/v5/mqtt5_client.c
```diff
@@ -590,6 +590,52 @@ static void s_enqueue_operation_front(struct aws_mqtt5_client *client, struct aw
...
+    if (aws_hash_table_put(manual_puback_cancelled_control_id_table, element->key, element->value, NULL)) {
+        return AWS_COMMON_HASH_TABLE_ITER_ERROR;
+    }
+
+    // We simply continue here and will clear (and thus decref) after we finish iterating.
+    return AWS_COMMON_HASH_TABLE_ITER_CONTINUE;
+}
+
+/* This is called when the manual puback entry is removed from a hashset to properly decref on removal */
+static void aws_mqtt5_manual_puback_entry_decref(void *value) {
```
sfod: nit: add the `s_` prefix

source/v5/mqtt5_client.c
```diff
@@ -3129,9 +3416,9 @@ int aws_mqtt5_client_service_operational_state(struct aws_mqtt5_client_operation
                  *  sporadically fail because the PINGRESP is processed before the write completion callback is
                  *  invoked.
                  *
-                 *  (2) Enqueue the ping - if the current operation is a large payload over a poor connection, it may
-                 *  be an arbitrarily long time before the current operation completes and the ping even has a chance
-                 *  to go out, meaning we will trigger a ping time out before it's even sent.
+                 *  (2) Enqueue the ping - if the current operation is a large payload over a poor connection, it
+                 * may be an arbitrarily long time before the current operation completes and the ping even has a
+                 * chance to go out, meaning we will trigger a ping time out before it's even sent.
```
sfod, line 3468: some unrelated changes creeped into PR (here and in couple other places)

include/aws/mqtt/private/v5/mqtt5_options_storage.h
sbSteveK: Moved both this and its paired destroy function into mqtt5_client.c

include/aws/mqtt/v5/mqtt5_client.h
```diff
@@ -345,6 +354,14 @@ struct aws_mqtt5_publish_completion_options {
     uint32_t ack_timeout_seconds_override;
 };
 
+/**
+ * Completion options for the manual puback operation
+ */
+struct aws_mqtt5_manual_puback_completion_options {
+    aws_mqtt5_manual_puback_completion_fn *completion_callback;
+    void *completion_user_data;
+};
```
sfod, line 364: Is it possible that a duplicate PUBLISH message is delivered to a client after the completion callback reported manual PUBACK for the previous instance of that PUBLISH as successful? For example, MQTT broker and client sent their packets simultaneously but client path was slightly faster.
If this scenario is possible, I believe the completion callback shouldn't be used at all because it'll give users a false confidence that they won't receive duplicates after PUBACK successfully completed.

source/v5/mqtt5_client.c
sbSteveK: I think these are formatting changes when I saved the file. They're deterministic when I tried to revert and save again.

include/aws/mqtt/v5/mqtt5_client.h
sbSteveK: I think we landed on the idea that the customer should be able to determine whether a publish is a duplicate whether or not they receive a duplicate publish. Unfortunately AWS IoT broker doesn't use the dup flag so there's no avoiding it. The pros of having a completion outweigh the cons but I'm open to discussion on it if you feel otherwise.

include/aws/mqtt/private/v5/mqtt5_client_impl.h
bretambrose: Following up on this:

I'd like to see the puback control an opaque subsystem with a clean/clear API.  Its usage within the 5 client would be similar to how the request response client embeds and uses the subscription manager.  

Advantages:

1. Separation of concerns - rather than dropping a bunch of new data structures into the 5 client impl and entangling the implementations, they stay independent
2. Testability - if we just dump everything into the 5 client then all testing requires a 5 client and a real connection to either a mock or real server.  If split off into an opaque subsystem with a clear API, the majority of the testing can be done without a client or an active connection; just use the "public"(private) subsystem API and introspect state
3. Reusability - if we need to add support to the 311 client, we don't need to reimplement everything from scratch; we just embed the subsystem instance into the 311 client and use its API correctly

sbSteveK: Keeping the manual puback operation in mqtt5_client.c feels like the cleaner way to handle this feature. The client PUBACK operation, manual or otherwise, feels like it belongs within the client as opposed to a separate subsystem. While it made sense to move it into a separate subsystem when originally considering both 3 and 5 clients, that's no longer the case. With everything already setup, it'd take the same amount of work to separate it now as it would to do so in the future if we ever add 3 support. There may also be thread or state related issues that we are not aware of right now that'd result in us redoing the work to support the 3 client that would render the current implementation unusable for the 3 client. I'd rather do that work when it's needed than unnecessarily do it if it's never needed.

From a testing perspective, I've included tests that use it in the same way we would want to test other client operations against a mock server rather. The mock server we have implemented is up to the task of detecting that PUBACKs are withheld and sent as expected. The test coverage feels like it's covering that the new API is doing what's expected, which with the removal of ordered PUBACKs is pretty simple.

Open to discussion and willing to change my mind but that's where I'm currently at.

bretambrose: Don't agree; dumping everything into one giant system is something we need to get away from.

source/v5/mqtt5_client.c
```diff
@@ -590,6 +590,53 @@ static void s_enqueue_operation_front(struct aws_mqtt5_client *client, struct aw
...
+/* This is used to move the aws_mqtt5_manual_puback_entry stored in the active table to cancelled table. We don't simply
+ * clear out the active table because we need to keep the entries in memory until we are fully done with them. */
+static int s_manual_puback_transfer(void *context, struct aws_hash_element *element) {
+    struct aws_hash_table *manual_puback_cancelled_control_id_table = context;
+
+    struct aws_mqtt5_manual_puback_entry *manual_puback_entry = element->value;
+    // incref the ref_count because when this entry is removed from the original set it will decref.
+    aws_ref_count_acquire(&manual_puback_entry->ref_count);
+    // Add the control id to the destination table
+    if (aws_hash_table_put(manual_puback_cancelled_control_id_table, element->key, element->value, NULL)) {
```
xiazhvera, line 600: Should we decref if the opreation failed?

sbSteveK: I don't think that can realistically fail in a way where everything isn't broken but I've moved the incref to AFTER the put operation

include/aws/mqtt/v5/mqtt5_client.h
sbSteveK: We are leaving this stuff here in case we want to add completion options in the future as well as to provide extra logging for debugging and info. We are however NOT binding the completion options out.

include/aws/mqtt/v5/mqtt5_client.h
```diff
@@ -738,6 +756,32 @@ int aws_mqtt5_client_publish(
     const struct aws_mqtt5_packet_publish_view *publish_options,
     const struct aws_mqtt5_publish_completion_options *completion_options);
 
+/**
+ * Takes manual control of PUBACK for the given PUBLISH packet.
+ *
```
xiazhvera, line 761: We should clarify that the function "should only ever be called by the user within the publish received callback" in the comments, otherwise the PUBACK will be sent after the publish callback is invoked. 

Also we should also mention that `0` indicates an invalid control.

source/v5/mqtt5_client.c
```diff
@@ -1883,20 +1939,26 @@ static void s_aws_mqtt5_client_connected_on_packet_received(
...
+                        AWS_LOGF_ERROR(
+                            AWS_LS_MQTT5_CLIENT,
+                            "id=%p: decode failure with error %d(%s)",
+                            (void *)client,
+                            error_code,
+                            aws_error_debug_str(error_code));
+
+                        s_aws_mqtt5_client_shutdown_channel(client, error_code);
+                    }
+                    return;
```
xiazhvera: trivial: the `return` seems unnecessary here.

source/v5/mqtt5_client.c
```diff
@@ -2430,6 +2492,243 @@ int aws_mqtt5_client_unsubscribe(
...
+    }
+
+    // The current_control_packet_id is incremented each time a new manual puback is scheduled.
+    uint64_t current_control_packet_id = client->operational_state.next_mqtt5_puback_control_id;
+    struct aws_mqtt5_manual_puback_entry *manual_puback =
+        s_aws_mqtt_manual_puback_entry_new(client->allocator, publish_view->packet_id, current_control_packet_id);
+
+    /* Allows lookup of manual puback entries by packet id. We incref here because the packet_id table also has a ref to
+     * the manual_puback */
+    aws_ref_count_acquire(&manual_puback->ref_count);
```
xiazhvera: Similarly, though it is unlikely to happen, if `aws_hash_table_put` failed, we need release the ref_count.

source/v5/mqtt5_client.c
```diff
@@ -2430,6 +2492,243 @@ int aws_mqtt5_client_unsubscribe(
...
+        goto cleanup;
+    }
+
+    /* Increment next_mqtt5_puback_control_id for next use */
+    client->operational_state.next_mqtt5_puback_control_id = current_control_packet_id + 1;
+
+    size_t in_flight_unacked_publishes =
+        aws_hash_table_get_entry_count(&client->operational_state.manual_puback_control_id_table);
+    if (in_flight_unacked_publishes >= 100) {
+        AWS_LOGF_WARN(
```
xiazhvera, line 2709: The in-flight message count here might not be accurate, as it may not account for in-flight messages that are waiting for PUBACK in the queue that were not acquired by the user.

xiazhvera (commented):
I temporary finished the main implementation part, still working on the tests...

source/v5/mqtt5_client.c
sbSteveK: Yea... This is a best-effort type of warn message. There are a number of things that can account for this number being inaccurate but we should at least log a warning related to potential issues rather than letting them go under the radar.

source/v5/mqtt5_client.c
sbSteveK: Moved the incref to after the put.

include/aws/mqtt/v5/mqtt5_client.h
```diff
@@ -294,6 +294,16 @@ typedef void(aws_mqtt5_publish_completion_fn)(
     int error_code,
     void *complete_ctx);
 
+/**
+ * Signature of callback invoked when a manual PUBACK operation completes.
+ *
+ * @param puback_result result of the PUBACK operation
+ * @param complete_ctx user data passed in with the completion options
```
sfod: super nit: incorrect var name

source/v5/mqtt5_client.c
```diff
@@ -1883,20 +1939,25 @@ static void s_aws_mqtt5_client_connected_on_packet_received(
...
-
-                    s_aws_mqtt5_client_shutdown_channel(client, error_code);
+                // Check if this PUBLSIH packet is manually controlled
+                struct aws_hash_element *elem = NULL;
+                aws_hash_table_find(
+                    &client->operational_state.manual_puback_packet_id_table, &publish_view->packet_id, &elem);
+
+                /* This PUBLISH isn't a manually controlled PUBACK. We schedule the PUBACK to be sent immediately. */
+                if (elem == NULL) {
+                    if (s_aws_mqtt5_client_queue_puback(client, publish_view->packet_id, NULL)) {
```
sfod, line 1950: Should we skip enqueueing PUBACK for the cases when the manual PUBACK with the same packet ID is already in the queue? Otherwise the queue may end up with two PUBACKs with the same packet ID. And since the MQTT broker can reuse the acked IDs, it can in theory send a new PUBLISH with the same ID between the first and the second PUBACKs. And the client will ack this new PUBLISH.

include/aws/mqtt/v5/mqtt5_client.h
sfod: trivial: `complete_ctx` should be renamed to `completion_user_data`

source/v5/mqtt5_client.c
sbSteveK: Discussed offline. We decided to leave this here as this scenario or a variant of it is possible across any point due to the nature of the redriven PUBLISH packets from the broker.

sfod (approved):

PR #422, opened 2026-04-23
Metrics metadata

xiazhvera:
*Issue #, if available:*

*Description of changes:*


source/v5/mqtt5_options_storage.c
```diff
@@ -699,17 +701,18 @@ int aws_mqtt5_packet_connect_storage_init(
         return AWS_OP_ERR;
     }
 
-    if (view->username != NULL) {
-        storage->username = *view->username;
+    if (view->username != NULL || (client_options_storage && client_options_storage->metrics_storage != NULL)) {
```
bretambrose: We do this twice; let's turn it into a simple one-line function ala 's_has_non_empty_username`

bretambrose (approved):
Didn't look at the tests too closely, but let's make sure there's a test that has a null username and still generates a metrics string given the gap that was there until this PR.

PR #423, opened 2026-04-24
Refactor IoT metrics parsing

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Refactor the metrics build functions, so it now handle duplicate params. If the username already have a param that the same as the metrics key, then we will not append the metrics (except for metadata field.)
- If the username already has a Metadata field, the library will append the metadata entries into the Metadata field.  

Step-by-step process:
 1. Parse existing username: Find the last '?' in the original username and parse query params into
 username_params_list
 2. Add "SDK" and "Platform" parameters if not already present
 3. Handle Metadata entries if metadata entries are provided
  a. Check if existing "Metadata" parameter exists in `username_params_list`
  b. If existing Metadata found with valid format:
     - Parse existing entries into a key-value `metadata_param_list`
     - Remove existing Metadata from `username_params_list` (will be rebuilt later)
  
    c. If existing Metadata has an invalid format:
      - Log debug message and skip adding metrics metadata entries
      - Keep original Metadata value unchanged, skip (d)
  
    d. Append metrics metadata entries to `metadata_param_list`
       - Add new metadata entries into `metadata_param_list` (existing keys take precedence, won't be overwritten)
       - build metadata value string from `metadata_param_list`
       - Add Metadata parameter to `username_params_list`
 4. Build final username from `username_params_list`


source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+static struct aws_uri_param *s_find_param_by_key(
+    struct aws_array_list *params_list,
+    const struct aws_byte_cursor *key,
+    size_t *out_index) {
+
+    size_t params_count = aws_array_list_length(params_list);
+    for (size_t i = 0; i < params_count; ++i) {
+        struct aws_uri_param *param = NULL;
+        aws_array_list_get_at_ptr(params_list, (void **)&param, i);
+        if (aws_byte_cursor_eq(&param->key, key)) {
```
bretambrose: What is our policy on case-sensitivity?

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
  */
-int s_build_username_query(
-    const struct aws_byte_cursor *base_username,
-    size_t base_username_length,
-    const struct aws_array_list *params_list,
-    struct aws_byte_buf *output_username,
-    size_t *out_final_username_size) {
+static int s_parse_delimited_entries(
+    const struct aws_byte_cursor content,
+    const struct aws_byte_cursor *delimiter,
```
bretambrose: Why is this a pointer and not required + positive length?

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
- * @param output_username [Optional] Buffer to write result. Caller must init/cleanup the buffer.
- * @param out_final_username_size [Optional] Outputs the final username size
+ * @param params_list List of aws_uri_param to search
+ * @param key The key to search for
+ * @param out_index [Optional] If not NULL and key is found, will be set to the index of the found parameter
+ * @return Pointer to the found parameter, or NULL if not found
+ */
+static struct aws_uri_param *s_find_param_by_key(
+    struct aws_array_list *params_list,
+    const struct aws_byte_cursor *key,
```
bretambrose: prefer pass-by-value on byte cursors that aren't output parameters

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+
+    struct aws_byte_cursor equals_delim = aws_byte_cursor_from_c_str("=");
+    struct aws_byte_cursor working_content = content;
+
+    while (working_content.len > 0) {
+        struct aws_byte_cursor entry_cursor;
+        struct aws_byte_cursor found_pos;
+
+        /* Find the next delimiter */
+        if (aws_byte_cursor_find_exact(&working_content, delimiter, &found_pos) == AWS_OP_SUCCESS) {
```
bretambrose: feels like this could be simplified and made less error-prone by using existing APIs like `aws_byte_cursor_next_split_on_cursor` or similar.

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+            entry_cursor.ptr = working_content.ptr;
```
bretambrose: prefer `aws_byte_cursor_advance` to raw arithmetic where appropriate.

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+            entry_cursor.len = found_pos.ptr - working_content.ptr;
+            working_content.ptr = found_pos.ptr + delimiter->len;
+            working_content.len = working_content.len - entry_cursor.len - delimiter->len;
+        } else {
+            /* No more delimiters, take the rest */
```
bretambrose: `aws_byte_cursor_advance`

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
 
-    AWS_ASSERT(base_username);
-    AWS_ASSERT(params_list);
+        /* Parse key=value from entry_cursor */
+        if (entry_cursor.len > 0) {
+            struct aws_uri_param entry_param;
+            AWS_ZERO_STRUCT(entry_param);
+
+            struct aws_byte_cursor equals_pos;
+            if (aws_byte_cursor_find_exact(&entry_cursor, &equals_delim, &equals_pos) == AWS_OP_SUCCESS) {
```
bretambrose: prefer split and/or use advance for the cursor manipulation

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+ * @param entry_delimiter Delimiter between entries (e.g., "&" for query strings, ";" for metadata)
+ * @param output_buf [Optional] Buffer to write the result to. Must be initialized with sufficient capacity.
+ * @param out_size [Optional] If not NULL, will be set to the size of the result string
+ * @return AWS_OP_SUCCESS on success, AWS_OP_ERR on failure
+ */
+static int s_build_delimited_params(
+    const struct aws_array_list *params_list,
+    const struct aws_byte_cursor *prefix,
+    const struct aws_byte_cursor *suffix,
+    const struct aws_byte_cursor *entry_delimiter,
```
bretambrose: This isn't optional, I would pass by value

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+ * @param prefix [Optional] String to prepend (e.g., "?" for query strings, "(" for metadata)
+ * @param suffix [Optional] String to append (e.g., ")" for metadata)
+ * @param entry_delimiter Delimiter between entries (e.g., "&" for query strings, ";" for metadata)
+ * @param output_buf [Optional] Buffer to write the result to. Must be initialized with sufficient capacity.
+ * @param out_size [Optional] If not NULL, will be set to the size of the result string
+ * @return AWS_OP_SUCCESS on success, AWS_OP_ERR on failure
+ */
+static int s_build_delimited_params(
+    const struct aws_array_list *params_list,
+    const struct aws_byte_cursor *prefix,
```
bretambrose, line 159: Rather than making these pointers, why not by-value and pass in zero-length cursor to skip?  That way you don't have to check pointer nullity?

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+
+    /* If params_list is empty, return empty string (no prefix/suffix) */
+    if (params_count == 0) {
+        if (out_size) {
+            *out_size = 0;
+        }
+        return AWS_OP_SUCCESS;
+    }
+
+    if (out_size) {
```
bretambrose: Consider always calculate out_size (to a local) and then conditionally assign it at the end if out_size is set.  Then you don't have to always check non-null every time you add to it.

Combined with previous suggestion, this if-block becomes:

```
size_t local_out = 0;
local_out += prefix.len + suffix.len;
```

source/mqtt_iot_metrics.c
```diff
@@ -15,85 +15,335 @@
...
+            }
+            if (out_size) {
+                *out_size += entry_delimiter->len;
             }
         }
 
-        if (out_final_username_size) {
-            *out_final_username_size += 1;
+        /* Append key */
+        if (output_buf && param.key.len > 0) {
```
bretambrose, line 209: When would the key be zero-length?  Is that invalid?

xiazhvera: Yes, a key value pair such as "=value" is a valid input, though it will be treated as "value" eventually.

source/mqtt_iot_metrics.c
xiazhvera: I prefer keeping these as pointers to maintain consistency with the aws-c-* API conventions, where optional parameters are represented as pointers.

source/mqtt_iot_metrics.c
xiazhvera: We uses case-sensitive string comparison (`aws_byte_cursor_eq`) when comparing keys. This means "SDK" and "sdk" would be treated as different keys.

source/mqtt_iot_metrics.c
```diff
@@ -28,15 +28,15 @@ const size_t DEFAULT_QUERY_PARAM_COUNT = 10;
  * @return Pointer to the found parameter, or NULL if not found
  */
 static struct aws_uri_param *s_find_param_by_key(
-    struct aws_array_list *params_list,
-    const struct aws_byte_cursor *key,
+    const struct aws_array_list params_list,
```
bretambrose: params list should stay by pointer.  It's only byte cursor that we'd pass by value since it doesn't own anything.

source/mqtt_iot_metrics.c
```diff
@@ -76,52 +76,43 @@ static bool s_add_param_if_not_exists(
...
-                entry_param.key = entry_cursor;
-            }
-
-            aws_array_list_push_back(out_entries, &entry_param);
+        struct aws_uri_param entry_param;
+        AWS_ZERO_STRUCT(entry_param);
+
+        struct aws_byte_cursor equals_pos;
+        if (aws_byte_cursor_find_exact(&entry_cursor, &equals_delim, &equals_pos) == AWS_OP_SUCCESS) {
+            entry_param.key.ptr = entry_cursor.ptr;
```
bretambrose:
```
size_t equals_offset = equals_pos.ptr - entry_cursor.ptr;
entry_param.key = aws_byte_cursor_advance(&entry_cursor, equals_offset);
aws_byte_cursor_advance(&entry_cursor, 1);
```

feels simpler and less "raw" (uses verified API functions and not raw pointer assignments)

source/mqtt_iot_metrics.c
```diff
@@ -76,52 +76,43 @@ static bool s_add_param_if_not_exists(
...
+            entry_param.key.len = equals_pos.ptr - entry_cursor.ptr;
+            aws_byte_cursor_advance(&entry_cursor, entry_param.key.len + 1);
+            entry_param.value.ptr = entry_cursor.ptr;
```
bretambrose: how does `entry_param.value.len` get set?  Should this be `entry_value.value = entry_cursor`?

source/mqtt_iot_metrics.c
```diff
@@ -161,16 +152,19 @@ static void s_merge_metadata_entries_no_overwrite(
  * @return AWS_OP_SUCCESS on success, AWS_OP_ERR on failure
  */
 static int s_build_delimited_params(
-    const struct aws_array_list *params_list,
+    const struct aws_array_list params_list,
```
bretambrose: Same here, revert to by ref

source/mqtt_iot_metrics.c
```diff
@@ -256,15 +245,15 @@ static int s_build_delimited_params(
  * @return AWS_OP_SUCCESS on success, AWS_OP_ERR on failure
  */
 static int s_build_metadata_value(
-    const struct aws_array_list *entries,
+    const struct aws_array_list entries,
```
bretambrose: Here

source/mqtt_iot_metrics.c
```diff
@@ -279,18 +268,15 @@ static int s_build_metadata_value(
  * @return AWS_OP_SUCCESS on success, AWS_OP_ERR on failure
  */
 static int s_build_username_query(
-    const struct aws_byte_cursor *base_username,
-    size_t base_username_length,
-    const struct aws_array_list *params_list,
+    const struct aws_byte_cursor base_username,
+    const size_t base_username_length,
+    const struct aws_array_list params_list,
```
bretambrose: And here

bretambrose (approved):

PR #431, opened 2026-07-29
Handle excess reason codes in SUBACK

sfod:
*Issue #, if available:* P480737743

*Description of changes:*
Use heap for SUBACK reason codes.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #432, opened 2026-08-05
Validate SUBACK/UNSUBACK reason code count

sfod:
This is a follow-up for https://github.com/awslabs/aws-c-mqtt/pull/431

Validate SUBACK/UNSUBACK reason code count against the acknowledged request. When a mismatch detected, the client fails the offending operation and disconnects sending a DISCONNECT with a reason code.

### DIFF
--- a/source/v5/mqtt5_client.c
+++ b/source/v5/mqtt5_client.c
@@ -3512,6 +3512,28 @@ int aws_mqtt5_client_service_operational_state(struct aws_mqtt5_client_operation
     return AWS_OP_SUCCESS;
 }
 
+static bool s_aws_mqtt5_client_validate_ack_reason_code_count(
+    const struct aws_mqtt5_operation *operation,
+    enum aws_mqtt5_packet_type packet_type,
+    const void *packet_view) {
+
+    if (packet_type == AWS_MQTT5_PT_SUBACK) {
+        const struct aws_mqtt5_packet_subscribe_view *subscribe_view = operation->packet_view;
+        const struct aws_mqtt5_packet_suback_view *suback_view = packet_view;
+
+        return subscribe_view->subscription_count == suback_view->reason_code_count;
+    }
+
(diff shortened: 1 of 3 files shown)


source/v5/mqtt5_client.c
```diff
@@ -3530,23 +3554,45 @@ void aws_mqtt5_client_operational_state_handle_ack(
...
-    s_complete_operation(client_operational_state->client, operation, error_code, packet_type, packet_view);
+    if (!s_aws_mqtt5_client_validate_ack_reason_code_count(operation, packet_type, packet_view)) {
+        AWS_LOGF_ERROR(
+            AWS_LS_MQTT5_CLIENT,
+            "id=%p: received a %s with a reason code count that does not match the acknowledged request; treating as a "
+            "protocol error and disconnecting",
+            (void *)client,
+            aws_mqtt5_packet_type_to_c_string(packet_type));
+
+        /*
```
bretambrose, line 3577: Do we do anything similar elsewhere in the file?  Should we move this (or just the shutdown block) into a helper function?

sfod: There's only one other place, in the `s_process_read_message` function, where the channel is shut down with a similar logic. Though it additionally checks for `error_code`:
```c
if (error_code == AWS_ERROR_MQTT5_DECODE_PROTOCOL_ERROR && s_should_client_disconnect_cleanly(client)) {
    s_aws_mqtt5_client_shutdown_channel_clean(client, error_code, AWS_MQTT5_DRC_PROTOCOL_ERROR);
} else {
    s_aws_mqtt5_client_shutdown_channel(client, error_code);
}
```
Checking the `error_code` doesn't fit well here and I'm not sure it'll make sense in other places where we could add it. So I don't see value in extracting the shutting down logic into a separate function.

As for `s_complete_operation`, AFAICS it's needed only here because at this point the operation is not in the operations list. Everywhere else it is.

bretambrose (approved):

PR #435, opened 2026-08-19
Resolve MSAN failure

sfod:
*Issue #, if available:*

One of the tests uses uninitialized memory as a payload, which triggers MSAN.

*Description of changes:*

While checking decoding random data is useful, it'll be hard to reproduce the issue in case of a failure. Instead, initialize payload buffer with pseudo-random data, which should cover everything covered by garbage.


bretambrose (approved):
A simple pattern is fine for the payload, since it's not really the focus of the test anyways.

